PR #6, opened 2016-03-12
readme: fix tabs -> spaces in step 2

cfcs:


talex5:
Spaces fix looks good, but author info is wrong:

    Author: root <root@localhost>

Not sure about the `-n` stuff. I think I tried it that way the first time and had some problems. We should just get all the dependencies released and remove the pins...

cfcs:
What's the problem with the author info?

`-n` stuff: Hmm, worked for me (tm), can you reproduce the errors?

- Getting stuff merged and removing pins: Sounds good! ;)

talex5:
`root@localhost` is not a useful address for recording who authored this change, or for contacting them in case of problems.

Anyway, another pin is gone now, and I fixed the indentation problem while updating that section...

cfcs:
I don't use email, so not sure what else I should put there.

Anyway, great! `-n` saved me a lot of time when building the second time around, but let's hope that that pins will eventually be unneeded either way :-)

Issue #18, opened 2017-12-02
how to use / configure this?


h01ger:
is this firewall configured exactly the same way as the "original" QubesOS one?

how does one configure the rules this one? are there any files which (for normal usage) can be edited besides rules.ml?

would you accept patches for README.md clarifying these questions? or would you prefer some other new file?

talex5:
The README currently says:

> Note: This firewall ignores the rules set in the Qubes GUI. See rules.ml for the actual policy.

This should be the only file that needs changing, but of course you might find a limitation of the firewall and need to extend the rules. Hopefully such extensions can be merged back here.

Qubes gives the firewall rules to the firewall as Linux iptables commands, which isn't very useful to us. There were some plans to change that, but I don't know what the outcome was:

https://groups.google.com/d/msg/qubes-devel/kvbKXT5CJMU/rSMD6Je3BgAJ

There is some documentation in the `rules.ml`, but feel free to clarify it. One thing that might be useful: replacing the polymophic variants with plain ones (e.g. `` `Nat`` -> `Nat`). That might give clearer error messages if the user makes a mistake.

h01ger:
I tried to modify rules.ml to allow connections from template VMs to 10.137.255.254, tcp, port 8082, but I didn't understand how. Could you please clarify how to achieve this, either by names or ip or mac addresses?

h01ger:
btw, you might also want to read http://layer-acht.org/thinking/blog/20171204-qubes-mirage-firewall/ - I'll post a link to the qubes-user list once the event here is over and I learned some more.

also: many thanks for qubes-mirage-firewall, I"m quite very happy with it already!

talex5:
Connections from template VMs should already be allowed by the default rules.

It would be nice to restrict templates to using *just* the update servers, though.

I think you can get the IP with something like:

```
| { src = `Client iface } when is_template iface#other_ip -> ...
```

But you have to write `is_template` yourself. It would be nice if the firewall could classify the VMs for you, or tag them with their names. Then you could do e.g.

```
| { src = `Client { name = "fedora-26-template" } } -> ...
```

Unfortunately, I don't think that Qubes provides this information to the firewall.

h01ger:
On Mon, Dec 04, 2017 at 04:48:26PM +0000, Thomas Leonard wrote:
> Connections from template VMs should already be allowed by the default rules.

not to 10.137.255.254, which is sys-net, serving as update-proxy for the
[...]

PR #19, opened 2017-12-02
explain how to build on Debian stretch too

h01ger:
Not everybody has docker easily available (eg ironically QubesOS users) so I figured it's useful to provide simple step-by-step instructions how to build this on a bare Debian system.

Signed-off-by: Holger Levsen <holger@layer-acht.org>

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ See [A Unikernel Firewall for QubesOS][] for more details.
 
 ## Build
 
+### with docker
+
 Clone this Git repository and run the `build-with-docker.sh` script:
 
     sudo yum install docker
@@ -19,8 +21,24 @@ Clone this Git repository and run the `build-with-docker.sh` script:
 
 This took about 10 minutes on my laptop (it will be much quicker if you run it again).
 
+### general build instructions
+
 You can also build without Docker, as for any normal Mirage unikernel;
 see [the Mirage installation instructions](https://mirage.io/wiki/install) for details.
+
+### with Debian stretch/stable
+
+    sudo apt install git ocaml-native-compilers camlp4-extra opam aspcud curl debianutils m4 ncurses-dev perl pkg-config time
+    git clone https://github.com/talex5/qubes-mirage-firewall
+    cd qubes-mirage-firewall/
+    opam init
+    opam switch 4.04.2
+    eval `opam config env`
+    ## in there:
+    opam install -y vchan xen-gnt mirage-xen-ocaml mirage-xen-minios io-page mirage-xen mirage mirage-nat mirage-qubes netchannel
+    opam config exec -- mirage configure -t xen
+    make depend
+    opam config exec -- make tar
 
 ## Deploy
 


talex5:
I removed the non-Docker instructions because it was too hard to keep them up-to-date:

https://github.com/talex5/qubes-mirage-firewall/commit/583366b22bf78be060bf3ab7be353bc7a3479f79#diff-04c6e90faac2675aa89e2176d2eec7d8

It's much easier to debug build problems if everyone is using the same versions and less confusing for users to have one set of instructions. Docker is easily available in Qubes (the README shows how to install it), although you might need to make its cache persistent if you're not using using a standalone VM for building (otherwise I guess you'll need to re-download the base image after rebooting).

(example of user problem: http://roscidus.com/blog/blog/2016/01/01/a-unikernel-firewall-for-qubesos/#comment-3090627184)

Of course, people are welcome to install without Docker. But I think only mirage devs are going to want to do that, and they'll already know how to do it.

h01ger:
Hi Thomas,

On Sun, Dec 03, 2017 at 02:53:43PM +0000, Thomas Leonard wrote:
> I removed the non-Docker instructions because it was too hard to keep them up-to-date:

thanks for explaining. (& fair enough!)

> It's much easier to debug build problems if everyone is using the same versions 
> and less confusing for users to have one set of instructions. 

true.

> Docker is easily available in Qubes (the README shows how to install it),
> although you might need to make its cache persistent if you're not using
> using a standalone VM for building (otherwise I guess you'll need to
> re-download the base image after rebooting).

see, I wouldn't call that easy.

> Of course, people are welcome to install without Docker. But I think only
> mirage devs are going to want to do that, and they'll already know how to do it.

I disagree & I'm not a mirage developer :) But you probably have a point
that this can also be fixed by improving the documentation on the mirage
side.

And last not least, I've solved it for me, so this works for me! ;)

In related news I plan to try qubes-mirage-firewall shortly and will
then try to come up with a patch documentating it's usage and
configuration. (issue #18)


-- 
cheers,
	Holger

h01ger:
On Sun, Dec 03, 2017 at 07:07:00AM -0800, Holger Levsen wrote:
> > Docker is easily available in Qubes (the README shows how to install it),
> > although you might need to make its cache persistent if you're not using
> > using a standalone VM for building (otherwise I guess you'll need to
> > re-download the base image after rebooting).
> see, I wouldn't call that easy.

that alone maybe could be described as easy, but with Debian Qubes you also
need to install the kernel headers from dom0 or run the Qube with its own
Debian kernel, and those two bits are not easy in my book.


-- 
cheers,
	Holger

talex5:
I don't recall having to do anything like that (I'm running Docker in a standalone Debian Qube).

Looking at my config, I see I'm using `Storage Driver: btrfs`, where `/var/lib/docker` is a btrfs filesystem. But I don't think that's necessary. I just tried running it with `sudo dockerd --data-root=/tmp/docker` and that worked too (it automatically used `Storage Driver: overlay2`).

The kernel is

```
$ uname -a
Linux dev 4.9.56-21.pvops.qubes.x86_64 #1 SMP Tue Oct 17 23:58:50 UTC 2017 x86_64 GNU/Linux
```

h01ger:
On Sun, Dec 03, 2017 at 04:51:49PM +0000, Thomas Leonard wrote:
> I don't recall having to do anything like that (I'm running Docker in a standalone Debian Qube).

I think it would be more productive to see whether we can get the same
checksum for a build with docker and without. AFAwe know (from tests
with reprotest), qubes-mirage-firewalls builds reproducible indepent
of most standard variations except the build path (date, time, locale,
etc doesnt matter and can vary..), provided the same ocaml version and
depends obviously. so my current question is: in which path does docker
build?


-- 
cheers,
	Holger

talex5:
The version of gcc also affects it, IIRC. It changed after a minor update to Debian once at least.

talex5:
(and docker builds it in `/home/opam/qubes-mirage-firewall` - see the Dockerfile for the exact commands it is running)

h01ger:
On Mon, Dec 04, 2017 at 07:59:04PM +0000, Thomas Leonard wrote:
> The version of gcc also affects it, IIRC. It changed after a minor update to Debian once at least.

yes, sure. what I ment with variations in this context was the
environment *besides* the exact same toolchain.

in build-with-docker.sh there is a SHA2 hash, which is for the unikernel
and not for the qubes taylored tarball, right? cause the tarball shows 
timestamp variations for the files in it...

btw, do you have plans/blockers to update to a debian 9 docker template? Maybe
I can reproduce that SHA2 when building natively? ;)

Thanks!


-- 
cheers,
	Holger

talex5:
It should be possible to make the tarball generation reproducably too. The other files inside it are just dummy placeholders to make Qubes happy. Ideally they wouldn't be there.

You should be able to test on any version of Debian just by changing the header of the Dockerfile.

h01ger:
On Thu, Dec 07, 2017 at 02:37:24AM -0800, Thomas Leonard wrote:
> It should be possible to make the tarball generation reproducably too. 

sure. tar just need to be told to use SOURCE_DATE_EPOCH for the mtimes
of the tar members. if building from git, setting S_D_E to the date of
the last commit is easy, only if not building from git, setting S_D_E
is more difficult, though setting it to 0 in those cases is an easy way
out.

https://reproducible-builds.org/specs/source-date-epoch/

> The other files inside it are just dummy placeholders to make Qubes happy. Ideally they wouldn't be there.

sure, but they are there.

> You should be able to test on any version of Debian just by changing the header of the Dockerfile.

cool.


-- 
cheers,
	Holger

PR #26, opened 2018-01-06
add qubes 4 hints

yomimono:


yomimono:
Instructions are missing for setting the VM type; Qubes 4 RC3 requires the user to specify that this should be PV rather than HVM.  I don't have an RC3 machine handy but I think @reynir has notes on this.

yomimono:
Superseded by #27.

PR #27, opened 2018-01-06
Add installation instructions for Qubes 4

### LINKED ISSUE #23: Must set virt_mode to 'pv' on Qubes R4
On Qubes R4-rc3 I had to set the `virt_mode` to `pv` instead of `hvm` for the mirage-firewall qube/vm.

### LINKED ISSUE #25: Is it supposed to work in Qubes 4?
I tried to follow the steps from the readme to install the binary release.

One thing to note is that the create new VM does not have ProxyVM. So I picked standalone not from template.

When I try to launch it with qvm-start, I briefly get a window showing booting from rom, which then closes and the vm with it.

The log files do show "cannot connect to qrexec" "cannot connect to gui-daemon" "cannot connect to qubesdb".

talex5:
Untested, but based on messages from Ralph, @yomimono and @reynir.

Fixes #23.
Fixes #25.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,12 @@ Note: This firewall *ignores the rules set in the Qubes GUI*. See `rules.ml` for
 
 See [A Unikernel Firewall for QubesOS][] for more details.
 
-## Build
+
+## Binary releases
+
+Pre-built binaries are available from the [releases page][].
+
+## Build from source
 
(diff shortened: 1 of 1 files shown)


yomimono:
I gave these instructions a quick run-through on my Qubes 4 machine and they seem to work.

PR #28, opened 2018-03-12
Don't emit TODO into console output every 5 minutes

h01ger:
It's indeed desirable to fix this, as to have correct timestamps in logs/console output, but still not every user needs to be reminded every 5 minutes.

### DIFF
--- a/command.ml
+++ b/command.ml
@@ -13,7 +13,7 @@ module Log = (val Logs.src_log src : Logs.LOG)
 let set_date_time flow =
   Flow.read_line flow >|= function
   | `Eof -> Log.warn (fun f -> f "EOF reading time from dom0"); 1
-  | `Ok line -> Log.info (fun f -> f "TODO: set time to %S" line); 0
+  | `Ok line -> 0 (* TODO: set time to "%S" line *)
 
 let handler ~user:_ cmd flow =
   (* Write a message to the client and return an exit status of 1. *)


h01ger:
how can the travis CI build be retriggered?

talex5:
If we don't log the timestamp, then how can the user know what the times in the log really are?

h01ger:
about the actual commit at hand, is there any problem with just merging it? shall I maybe resubmit with a minor whitespace change? 

I understand that dealing with time properly is a major undertaking, but this is just a very simple commit cleaning up the user visible console log from showing a development todo reminder every 5 minutes...

talex5:
But I need the timestamps to know what the time is. At least currently, Qubes dom0 doesn't add its own to the logs.

h01ger:
Thomas, there are timestamps now. This commit is about not emitting an entry which says "TODO: set the time to $currenttime+-2secs" every 5min. This commit is NOT about removing timestamps from the logs.

Sigh. I'm getting really frustrated about this now. This clearly is not the greatest enhancement in the world, but the feedback here feels really discouraging and I'm not sure I can recommend contributing to q-m-f cause my experiences were always frustrating.

talex5:
Without the TODO line, the other timestamps aren't useful. I'm not at a Qubes machine right now, but as I recall the logs look something like:

```
02:00 Some message
03:00 The time is really 05:00 now!
04:00 Something that really happened at 06:00.
```

If you don't log the actual time, you can't work out what the other timestamps mean. I'd be happy to accept a PR that actually fixes the problem though.

It sounds like you have some syslog solution planned for this, but at least whenever I've used the Qubes GUI to see the logs, the only timestamps I've seen are the ones added by the unikernel itself.

h01ger:
thanks for that explaination, this makes sense.

PR #47, opened 2019-04-03
Update dependencies

talex5:
Remove pin on mirage 3.4 - it should now be working with the latest release (untested - @xaki23: does this work for you now?).


xaki23:
qubes-builder build works, and resulting template runs fine for a quick functionality check.
thanks for the update!

PR #49, opened 2019-04-05
Use source date in .tar.bz2 archive

talex5:
All files are now added using the date the build-with-docker script was last changed. Since this includes the hash of the result, it should be up-to-date. This ensures that rebuilding the archive doesn't change it in any way.

Fixes #19


yomimono:
Thanks!

PR #52, opened 2019-04-08
Add patch to cmdliner for reproducible build

talex5:
See https://github.com/dbuenzli/cmdliner/pull/106


yomimono:
Seems fine; we should track whether this gets merged upstream.

PR #55, opened 2019-04-16
Upgrade to latest mirage-nat to fix ICMP

### LINKED ISSUE #50: an upstream tcpip bug breaks ping in latest release
The latest `qubes-mirage-firewall` release breaks `ping` from downstream clients, as outbound ICMP echo requests translated by the firewall get an invalid ICMP checksum and are not answered by popular remote servers like 8.8.8.8 .  This isn't the firewall's fault, but once the upstream problem in `tcpip` is fixed and a new version released, the new version should be used to fix this regression.

talex5:
Now ping and traceroute work - thanks @yomimono!

Fixes #50.


yomimono:
Thanks also to my coauthor @linse and to @xaki23 for the initial report that lead to this work. :)

talex5:
(taking the above comment as approval)

PR #60, opened 2019-04-28
 Wait if dom0 is slow to set the network configuration

talex5:
If the firewall boots too quickly, it seems it can fail with:

```
ERR [application] main: (Failure "QubesDB key \"/qubes-ip\" not present")
```

A similar thing also happens if you haven't configured a NetVM, or didn't configure the firewall as a proxy. Now we just log a warning and wait until things improve.

Also, I updated the mirage-nat minimum version to make sure `ping` works.

### DIFF
--- a/config.ml
+++ b/config.ml
@@ -31,7 +31,7 @@ let main =
       package "mirage-net-xen";
       package "ipaddr" ~min:"3.0.0";
       package "mirage-qubes";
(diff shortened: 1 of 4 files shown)


yomimono:
I don't have a nice way to induce the error condition this fixes, but this looks correct and works for me.

PR #69, opened 2019-05-31
Fix ln(1) call in build instructions

jaseg:
The arguments were backwards. [```ln``` takes the link target first, then the link name](https://linux.die.net/man/1/ln).


talex5:
Thanks!

PR #74, opened 2019-07-28
minor updates

xaki23:
just a bundle of minor updates/fixes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ FROM ocaml/opam2@sha256:74fb6e30a95e1569db755b3c061970a8270dfc281c4e69bffe2cf990
 # Pin last known-good version for reproducible builds.
 # Remove this line (and the base image pin above) if you want to test with the
 # latest versions.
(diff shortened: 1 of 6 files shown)


xaki23:
interesting. 

the failure is because ci pins down the libs used (like the docker build), but doesnt pin down the q-m-f version (which makes sense, else it wouldnt be useful at all).
so it fails on the bytes/octets thing (for ipaddr-4.0.0) since it uses ipaddr-3.1.0 ...

i would prefer if someone who understands the docker build path updates the ipaddr version there over me pinning it down for the qubes-builder build.

xaki23:
note-to-future-self on this:

a) figure out the head commit hash of https://github.com/ocaml/opam-repository
b) stick that hash in the git-reset-hard line in Dockerfile
c) run a docker build
d) stick the this-build hash from the end of the build process in the last-known line in build-with-docker.sh
e) do a full rebuild in a clean dvm, just to be sure. the hashes should match now.

talex5:
Merged as #75 - thanks!

PR #75, opened 2019-07-28
Upstream updates

talex5:
A slightly modified version of #74:

- Squashed ipaddr changes into one commit.
- Capitalised log messages, for consistency with other commits.
- Modified Dockerfile to avoid installing and then downgrading packages.
- Modified Travis script to show build's sha256.

@xaki23 : I skipped the README change for now. It's probably correct, but I didn't understand the commit message:

> actualy create the symlink-redirected docker dir so the installer wont remove the dangling symlink

What installer is this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ script:
   - echo 'ADD . /home/opam/qubes-mirage-firewall' >> Dockerfile
   - echo 'RUN sudo chown -R opam /home/opam/qubes-mirage-firewall' >> Dockerfile
   - docker build -t qubes-mirage-firewall .
-  - docker run --rm -i qubes-mirage-firewall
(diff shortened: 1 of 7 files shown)


xaki23:
> @xaki23 : I skipped the README change for now. It's probably correct, but I didn't understand the commit message:
> > actualy create the symlink-redirected docker dir so the installer wont remove the dangling symlink
> What installer is this?

by installer i meant just dnf/rpm.
if you just set the symlink, but the target dir doesnt exist, the "dnf install docker" step will remove the symlink again and create /var/lib/docker as a dir.
if you "mkdir /home/user/docker" before "dnf install docker" (doesnt matter if before or after the ln) it actualy works as intended.

talex5:
Ah, so it does! I've added that change back in - thanks!

PR #76, opened 2019-08-05
Mirage-xen >= 4.0.0

reynir:
These changes are required for mirage-xen >= 4.0.0.

See also https://github.com/mirage/mirage/pull/995 since it's making mirage put a mirage-xen < 4.0.0 constraint.

PR #77, opened 2019-08-05
Add an upper bound for mirage-xen

reynir:
We use features from mirage-xen not in mirage-xen >= 4.0.0.

Related to #76, and https://github.com/mirage/mirage/pull/995 will likely break qubes-mirage-firewall.


talex5:
Thanks (sorry, I don't know why this got ignored). But closing in favour of #80, which instead updates to the new API.

PR #79, opened 2019-08-25
update to ocaml 4.08.1 and mirage-xen 4.x+ 

xaki23:
update to ocaml 4.08.1 and mirage-xen 4.x+


talex5:
Thanks! I've updated the Dockerfile too in #80.

Issue #35, opened 2018-11-04
mirage vm crash on downstream vm start/stop


xaki23:
somewhat outdated build, but build-with-docker failing currently for me, and i can not find any closed issue that sounds related, so may be unfixed yet.

mirage fw vm sometimes exits on downstream vm start/stop, with errors like this:

-----------------------------------------
2018-11-04 01:19:13 -00:00: INF [client_net] add client vif {domid=462;device_id=0}
Fatal error: exception Xs_protocol.Enoent("unwatch")
Raised at file "map.ml", line 122, characters 10-25
Called from file "src0/sexp_conv.ml", line 149, characters 10-37
Mirage exiting with status 2
Do_exit called!

xaki23:
afaict this still happens with a 20190416 build (aka "beyond the icmp issue").

i moved the backup vm (which gets restarted hourly) to a separate mirage-firewall instance some months ago, and it looks like that backup-mfwt is still exploding multiple times per week.

talex5:
I've made a test branch with a version of XenStore that logs this error instead of crashing.

To test:

```
git clone 'https://github.com/talex5/qubes-mirage-firewall.git' -b unwatch-crash
cd qubes-mirage-firewall/
./build-with-docker.sh
```

Then deploy as usual. On start-up, this version should log:

```
2019-11-22 11:36:15 -00:00: INF [xenstore.client] ** XenStore with unwatch patch **
```

If it still crashes, it should crash with a different error, showing the real problem. If it no longer crashes, please check the logs after a while for `ERR` messages.

xaki23:
built and deployed on two systems in a role that triggers this bug 2-3 times a week.
(fw for backup-sync vm that gets started/stopped once per hour)

xaki23:
2019-12-17 04:36:41 -00:00: INF [xenstore.client] ** XenStore with unwatch patch **
....
2019-12-19 17:33:22 -00:00: INF [net-xen backend] Frontend asked to close network device dom:3167/vif:0
2019-12-19 17:33:22 -00:00: INF [client_net] client {domid=3167;device_id=0} has gone
2019-12-19 18:25:56 -00:00: INF [client_net] add client vif {domid=3168;device_id=0} with IP 10.137.0.26
2019-12-19 18:25:56 -00:00: ERR [xenstore.client] Error unwatching "/local/domain/3168/device/vif/0/state": Xs_protocol.Enoent("unwatch")
[...]

PR #82, opened 2019-11-22
Use patched XenStore to debug unwatch crash

talex5:
This might be useful for debugging the cause of #35.


xaki23:
maybe close this one since we know the reason for #35 now, just dont have a fix yet?

talex5:
OK, closing.

PR #83, opened 2019-11-27
Note that AppVM Size may need to increase

marmot1791:
Add note that AppVM used to build from source may need a private image larger than the default 2048MB.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,9 +15,8 @@ See the [Deploy](#deploy) section below for installation instructions.
 
 ## Build from source
 
-Create a new Fedora-29 AppVM (or reuse an existing one). Open a terminal.
-Note that you may need more than the default 2GB (2048MB) of storage in the private 
-image of the AppVM, so you may need to increase the size in the Qube's Settings.
+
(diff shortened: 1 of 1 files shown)


talex5:
2GB used to be enough, but I just tried it on a Fedora 30 template and it does indeed fail now:

```
[ERROR] Could not update repository "default": "/bin/cp -PRp /home/opam/opam-repository /home/opam/.opam/repo/default.new" exited with code 1
The command '/bin/sh -c cd ~/opam-repository && git fetch origin master && git reset --hard 5eed470abc5c7991e448c9653698c03d6ea146d1 && opam update' returned a non-zero code: 40
```

Looks like it's actually out of inodes rather than space (see `df -hi`), but increasing the space does fix it.

The instructions should say how much space will be needed. Maybe something like this?

> Create a new Fedora-30 AppVM (or reuse an existing one). In the Qube's Settings (Basic / Disk storage), increase the private storage from the default 2048 MiB to 4096 MiB. Open a terminal.

marmot1791:
Updated per your comment.

talex5:
Thanks!

PR #89, opened 2020-01-11
support Mirage 3.7 and mirage-nat 2.0.0

### LINKED ISSUE #73: fragmented ip handling
it seems there are still some basic problems with the mfw IP stack.
todays variant: fragmented packets.

the tests are against a 20190727 (today) build in templated mode.
the test commands are run in a (fedora-29) vm below the mirage-vm.
the log lines are from mfw-console-logs in dom0.
i am not pasting the hex dumps because afaict they dont matter and are easy to reproduce if anyone cares...

$ traceroute -U 8.8.8.8 2000

2019-07-27 18:33:40 -00:00: WRN [client_net] Ignored unknown IPv4 message: Failed to parse UDP packet: UDP header claimed a payload longer than the supplied buffer: 1972 vs 1480.
2019-07-27 18:33:40 -00:00: WRN [client_net] Ignored unknown IPv4 message: Failed to parse UDP packet: UDP header claimed a payload longer than the supplied buffer: 17469 vs 500.

the second line is almost more scary, because that means it tries to parse the payload/data of the second fragment as a header.

when using icmp instead (traceroute with -I instead of -U or "ping -s 2000 8.8.8.8") the failure pattern is mostly the same, except it doesnt complain about the first fragment with either a 20190727 nor 20190226 (pre #31) build.

and proactively pinging @yomimono on this.

note this is the _other_ direction compared to #31, it already falls on its face while the pakets are coming from the inside.

hannesm:
The main improvement is fragmentation and reassembly support in mirage-nat 2.0.0, which fixes #73 //cc @xaki23.

Other changes include:
- update to mirage-xen 5.0.0 (which reverted the split OS -> OS & Os_xen, back to OS),
- upgrade to mirage-qubes 0.8.0 (where GUI.listen signature changed),
- fix deprecation warnings (Mirage_clock_lwt is deprecated, using Mirage_clock instead),
- remove Fw_utils.return and fail, use Lwt.return_unit across the codebase (instead of Lwt.return () or return ()).

Fragmentation and reassembly: each network interface now has a reassembly cache, a LRU, its size is hardcoded to 256 * 1024 bytes. I am not sure whether this should instead (a) be user-configurable (b) default to a smaller size. Maybe the individual client networks (Client_net) should have a smaller cache, and the Uplink a slightly larger?

### DIFF
--- a/firewall.ml
+++ b/firewall.ml
@@ -1,7 +1,6 @@
 (* Copyright (C) 2015, Thomas Leonard <thomas.leonard@unikernel.com>
    See the README file for details. *)
 
-open Fw_utils
 open Packet
 open Lwt.Infix
 
@@ -32,7 +31,7 @@ let transmit_ipv4 packet iface =
             Log.warn (fun f -> f "Failed to write packet to %a: %s"
                          Ipaddr.V4.pp iface#other_ip
                          (Printexc.to_string ex));
-            Lwt.return ()
+            Lwt.return_unit
          )
     )
     (fun ex ->
@@ -40,7 +39,7 @@ let transmit_ipv4 packet iface =
                    (Printexc.to_string ex)
                    Nat_packet.pp packet
                );
-       Lwt.return ()
+       Lwt.return_unit
     )
 
 let forward_ipv4 t packet =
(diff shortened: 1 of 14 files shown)


firewall.ml
```diff
@@ -15,29 +14,32 @@ let transmit_ipv4 packet iface =
...
          )
     )
     (fun ex ->
        Log.err (fun f -> f "Exception in transmit_ipv4: %s for:@.%a"
                    (Printexc.to_string ex)
                    Nat_packet.pp packet
                );
-       Lwt.return ()
+       Lwt.return_unit
     )
```
hannesm: I'm curious why there are two `Lwt.catch` here -- looking at the git history I stumbled across this commit https://github.com/mirage/qubes-mirage-firewall/commit/e070044fefcd43d6024099bf27794bfe474bec31?w=1 -- and I think one of the `Lwt.catch` blocks can be safely removed.

xaki23:
this is very confusing. i just built this...
 - from a clean buildvm (so no old ocaml/opam anything), 
 - with qubes-builder https://github.com/QubesOS/qubes-builder/blob/master/example-configs/mirage.conf
 - swapped out the repo/branch for roburio/mirage-3.7
 - swapped mirage pin in Makefile.builder to 3.7.4 (as in Dockerfile)
 - build finished within <10 min _including_ "have to install opam + ocaml + a lot of modules first"
 - didnt spew the usual bulk text about actualy building firewall itself (that looks very different from opam-is-installing-modules)
 - got a reasonable looking ~5-6MB mirage-firewall.xen
 - templated + installed + set it to be used for my backup sync process ... 

... and it just seems to work.

this is actualy so smooth i am not sure i am looking at the right thing.
could we get some versions in the during-startup debugprints or so?
(yes, no timestamps because reprobuilds, but versions should be ok for that)

will deploy it to some more roles tomorrow and watch it for a few days.
i dont have strong active testing for this, but things tend to blow up quickly for me during use.

talex5:
@hannesm : looks great - thanks!
@xaki23 : the mirage tool now does a "quiet" build by default (https://github.com/mirage/mirage/pull/999)

firewall.ml
talex5: Looks like it was originally distinguishing between errors from `Nat_packet` and errors doing the write. But since 0852aa0f43 the outer catch will never be used. I've removed it in #90.

PR #90, opened 2020-01-13
Minor cleanups

talex5:
- Fix Lwt.4.5.0 in the Dockerfile for faster builds (avoids downgrading in next step)
- Remove unused Clock argument to Uplink

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ FROM ocurrent/opam@sha256:3f3ce7e577a94942c7f9c63cbdd1ecbfe0ea793f581f69047f3155
 # latest versions.
 RUN cd ~/opam-repository && git fetch origin master && git reset --hard d205c265cee9a86869259180fd2238da98370430 && opam update
 
-RUN opam depext -i -y mirage.3.7.4 lwt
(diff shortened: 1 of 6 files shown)


xaki23:
"ported" the "faster build" part to Makefile.builder, works awesome. 
5-6 min for the full builder-build.
how do i +1 this for merge?

workflow question ... do you want that additional change:
 - as a review/comment here, 
 - a PR against talex5:cleanup while this is still open or 
 - a PR against mirage:master after this is merged?

perhaps we can get this capsulated in a way that the "pre-deps" need to be updated only in one place in the future? 
put the "opam depext" call into a separate .sh and call that from both dockerfile + makefile?
that smells, but less than double updates perhaps?

talex5:
At some point mirage will move to building with dune. Depending on how it does that, it might solve the problem by letting us specify the version of mirage in the dune file.

talex5:
I've merge it now. Feel free to make a separate PR updating `Makefile.builder`.

xaki23:
> I've merge it now. Feel free to make a separate PR updating `Makefile.builder`.

https://github.com/mirage/qubes-mirage-firewall/pull/91

PR #91, opened 2020-01-14
support mirage-3.7 via qubes-builder

xaki23:
basicly just replicates two minors to the qubes-builder path:
 - the move to mirage-3.7 
 - and early lwt-4.5 to avoid version-yoyo
 - and replaces pin by depext (which makes the depext part cut-n-paste from the Dockerfile)

change is tested+working.


talex5:
Thanks!

PR #96, opened 2020-02-14
QubesOS 4.0 support

hannesm:
this is a rebased + squashed + improved version of #85, main features are to read (and listen for changes) firewall rules from QubesDB; and support of hostnames in firewall rules (using the Dns_client).

A test unikernel is available in the `test` subdirectory (depending on a slightly modified version of Alcotest).

This project was done by @linse and @yomimono in summer 2019.

PR #99, opened 2020-05-15
bump qubes-builder ocaml-version to 4.10.0 for gcc-10 compatibility

xaki23:
been testing with 4.10.0 in #96 the last several builds already.
pinging @hannesm for "please just hit merge".

(build works now, seems i managed to hit the time window between merge here and mirage-nat 2.2.1 getting merged into opam earlier)


hannesm:
thanks

PR #103, opened 2020-05-19
Handle other IP formats from xenstore.

linse:
Example: "10.137.0.18 fd09:24ef:3178::a19:11"
reported via https://twitter.com/t_grote/status/1262747002334408704


grote:
I can confirm that this fixes the issue with my Qubes installation.

PR #109, opened 2020-06-19
share host system opam cache with docker build

hannesm:
Co-Authored-By: hannes <hannes@mehnert.org>


talex5:
I don't think this makes any difference. The downloading happens before this in the `docker build` step. You could use `DOCKER_BUILDKIT=1` and the caching features described at https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/experimental.md though.

hannesm:
thanks for your reply @talex5, I trust your knowledge about docker and caching.

Issue #70, opened 2019-06-12
Remove dummy kernel files


talex5:
The mirage-firewall distribution includes dummy `modules.img` and `initramfs` files. These were needed on Qubes 3 for it to accept something as a kernel.

On Qubes 4 (according to https://www.qubes-os.org/doc/managing-vm-kernel/):

> All the files besides vmlinuz are optional.

We should remove these dummy files from the build and installation instructions.

marmot1791:
As of today on Qubes R4.0 with all updates dom0 still requires the dummy initramfs to start a mirage qube, but modules.img is not required.

marmot1791:
Opened question of whether this is a bug or a docs issue over at qubes-issues: https://github.com/QubesOS/qubes-issues/issues/5516

marmot1791:
This is partially blocked until Qubes R4.1. Until that is released `initramfs` is still required.

hannesm:
Since Qubes 4.0 is end of life now, IMHO it would be good to remove the initramfs file. This also means that our packaging and install instructions can be much simpler (we'll only distribute a single binary, and no need for `tar`).

Does this make sense @palainp?

palainp:
This should be right according to the documentation, but I was unable to boot without it some times algo :/

palainp:
This is still the case, I have the following:
```
Missing file: /var/lib/qubes/vm-kernels/mirage-fw/initramfs
```

palainp:
The existence of initramfs is checked in https://github.com/QubesOS/qubes-core-admin/blob/c177aacbe1afc35034826ef538c0e49ed40b6bc7/qubes/storage/kernels.py#L140
See: https://github.com/QubesOS/qubes-issues/issues/5516

hannesm:
Thanks @palainp for your investigations :)

palainp:
https://github.com/QubesOS/qubes-issues/issues/5516 has been resolved now. This will works with the next Qubes update, and we can remove that dummy file for the next release :)

PR #110, opened 2020-06-19
Do not run tar in dom0 (closes #84).

### LINKED ISSUE #84: Running tar in dom0 with domU input
The suggested instructions invoke tar in dom0 using untrusted input from either the domain where mirage firewall is built or the downloaded mirage firewall binary. This seems unwise.

While we're placing a lot of trust in whatever domain is used to build the mirage-firewall, I don't think mirage-firewall actually does anything else that allows it to control the execution of code in dom0. Shouldn't the correct procedure be to untar the 3 files in the builder/downloader domain and then copy them individually to dom0?

burghardt:
Do not run tar and bzip2 in dom0 to decompresses and extract archive
data created in, or downloaded to domU as any vulnerabilities in them
can compromise Qubes OS security model.

Instead of that run both tar and bzip2 in domU and copy unikernel to
dom0 as described in official Qubes documentation ["Copying from (and to)
dom0"](https://www.qubes-os.org/doc/copy-from-dom0/#copying-to-dom0).

Auxiliary files required to run unikernel in Qubes OS domU can be easily
created directly in dom0 using trusted tools available there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,12 +48,21 @@ However, it should still work fine.
 
 ## Deploy
 
-If you want to deploy manually, unpack `mirage-firewall.tar.bz2` in dom0, inside `/var/lib/qubes/vm-kernels/`. e.g. (if `dev` is the AppVM where you built it):
+If you want to deploy manually, unpack `mirage-firewall.tar.bz2` in domU. The tarball contains `vmlinuz`,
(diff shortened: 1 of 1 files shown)


talex5:
That's better - thanks!

(it's possible we could skip the `modules.img` too now, according to #70)

PR #116, opened 2020-08-19
Upgrade to Mirage 6 for solo5 PVH support

### LINKED ISSUE #29: HVM support
Dear @talex5, many thanks for your work with awesome firewall!

1. I can't start the firewall in HVM mode in QubesOS, do you plan to introduce support for this mode?

2. Also I have another question, don't want create more tickets. It's not really issue but...
I have attached three appVMs to one miragefirewallVM
So considering that I'm using a standard setup of firewall (your binary files from the github), can appvm1 communicate with appvm2 or appvm3? Or are all machines strictly isolated from each other?

talex5:
This PR updates the Dockerfile to collect together the various pins needed to test solo5 PVH support from e.g. https://github.com/mirage/mirage/issues/1184.

Once this works, it will fix #29.

For me, it boots but then doesn't respond further (e.g. to `qvm-run` commands). Attempting to start a client just times out (probably because it doesn't connect the new network device).

```
$ test-mirage qubes_firewall.xen mirage-firewall-staging
Waiting for 'Ready'... OK
Uploading 'qubes_firewall.xen' (7340600 bytes) to "mirage-firewall-staging"
Waiting for 'Booting'... OK
Connecting to mirage-firewall-staging console...
Solo5: Xen console: port 0x2, ring @0x00000000FEFFF000
Solo5: using E820 memory map
            |      ___|
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
Solo5: Bindings version v0.6.6-2-g20196872
Solo5: Memory map: 32 MB addressable:
Solo5:   reserved @ (0x0 - 0xfffff)
Solo5:       text @ (0x100000 - 0x2edfff)
Solo5:     rodata @ (0x2ee000 - 0x350fff)
Solo5:       data @ (0x351000 - 0x4b7fff)
Solo5:       heap >= 0x4b8000 < stack < 0x2000000
2020-08-19 13:42:26 -00:00: INF [qubes.rexec] waiting for client...
2020-08-19 13:42:26 -00:00: INF [qubes.gui] waiting for client...
2020-08-19 13:42:26 -00:00: INF [qubes.db] connecting to server...
2020-08-19 13:42:26 -00:00: INF [qubes.db] connected
2020-08-19 13:42:26 -00:00: INF [qubes.rexec] client connected, using protocol version 2
```

In dom0, I ran `gdbsx -a $DOMID 64 9999` and then:
```
[tal@dom0 ~]$ gdb /var/lib/qubes/vm-kernels/mirage-firewall-staging/vmlinuz 
(gdb) target remote :9999
Remote debugging using :9999
warning: remote target does not support file transfer, attempting to access files from local filesystem.
warning: Unable to find dynamic linker breakpoint function.
GDB will be unable to debug shared library initializers
and track explicitly loaded dynamic code.
mirage_xen_evtchn_block_domain (v_deadline=<optimized out>) at bindings/evtchn.c:343
343	bindings/evtchn.c: No such file or directory.
(gdb) bt
#0  mirage_xen_evtchn_block_domain (v_deadline=<optimized out>) at bindings/evtchn.c:343
#1  0x00000000001a2191 in camlOS__Main__aux_141 () at lib/main.ml:56
#2  0x000000000010bb05 in camlMain__entry () at main.ml:5
#3  0x0000000000107e69 in caml_program ()
#4  0x0000000000282df4 in caml_start_program ()
#5  0x000000000026596c in caml_startup_common (argv=argv@entry=0x499c90 <unused_argv>, pooling=<optimized out>, 
    pooling@entry=0) at startup_nat.c:162
#6  0x00000000002659ab in caml_startup_exn (argv=argv@entry=0x499c90 <unused_argv>) at startup_nat.c:172
#7  caml_startup (argv=argv@entry=0x499c90 <unused_argv>) at startup_nat.c:172
#8  0x000000000026564c in solo5_app_main (si=si@entry=0x49b010 <si>) at bindings/main.c:114
#9  0x00000000001011a4 in _start2 (arg=<optimized out>) at xen/start.c:80
#10 0x00000000001010a2 in _newstack () at xen/boot.S:154
#11 0x0000000000000000 in ?? ()
```

/cc @mato

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,10 @@ See the [Deploy](#deploy) section below for installation instructions.
 
 ## Build from source
 
+Note: The most reliable way to build is using Docker.
+Fedora 30 works well for this, but installing Docker on Fedora 31 or 32 is more difficult.
+Debian 10 also works, but you'll need to follow the instructions at [docker.com][debian-docker] to get Docker
+(don't use Debian's version).
 
 Create a new Fedora-30 AppVM (or reuse an existing one). In the Qube's Settings (Basic / Disk storage), increase the private storage max size from the default 2048 MiB to 4096 MiB. Open a terminal.
 
@@ -32,8 +36,6 @@ It gives Docker more disk space and avoids losing the Docker image cache when yo
 
 Note: the object files are stored in the `_build` directory to speed up incremental builds.
 If you change the dependencies, you will need to delete this directory before rebuilding.
-
-If you want to build on Debian, follow the instructions at [docker.com][debian-docker] to get Docker and then run `sudo ./build-with-docker.sh` as above.
 
 It's OK to install the Docker package in a template VM if you want it to remain
 after a reboot, but the build of the firewall itself should be done in a regular AppVM.
@@ -59,12 +61,11 @@ Copy `vmlinuz` to `/var/lib/qubes/vm-kernels/mirage-firewall` directory in dom0,
     [tal@dom0 ~]$ cd /var/lib/qubes/vm-kernels/mirage-firewall/
     [tal@dom0 mirage-firewall]$ qvm-run -p dev 'cat mirage-firewall/vmlinuz' > vmlinuz
 
-Finally create dummy files required by Qubes OS:
+Finally, create [a dummy file required by Qubes OS](https://github.com/QubesOS/qubes-issues/issues/5516):
 
-    [tal@dom0 mirage-firewall]$ touch modules.img
     [tal@dom0 mirage-firewall]$ gzip -n9 < /dev/null > initramfs
 
-Run this command in dom0 to create a `mirage-firewall` VM using the `mirage-firewall` kernel you added above:
+Run this command in dom0 to create a `mirage-firewall` VM using the `mirage-firewall` kernel you added above
 
 ```
 qvm-create \
@@ -75,15 +76,28 @@ qvm-create \
   --property netvm=sys-net \
   --property provides_network=True \
   --property vcpus=1 \
-  --property virt_mode=pv \
+  --property virt_mode=pvh \
   --label=green \
   --class StandaloneVM \
   mirage-firewall
 
 qvm-features mirage-firewall qubes-firewall 1
+qvm-features mirage-firewall no-default-kernelopts 1
 ```
 
+**Note**: for `virt_mode`, use `pv` instead of `pvh` for firewall versions before 0.8.
+
+## Upgrading
+
 To upgrade from an earlier release, just overwrite `/var/lib/qubes/vm-kernels/mirage-firewall/vmlinuz` with the new version and restart the firewall VM.
+
+If upgrading from a version before 0.8, you will also need to update a few options:
+
+```
+qvm-prefs mirage-firewall kernelopts ''
+qvm-prefs mirage-firewall virt_mode pvh
+qvm-features mirage-firewall no-default-kernelopts 1
+```
 
 ### Configure AppVMs to use it
 
--- a/memory_pressure.ml
+++ b/memory_pressure.ml
@@ -6,44 +6,48 @@ open Lwt
 let src = Logs.Src.create "memory_pressure" ~doc:"Memory pressure monitor"
 module Log = (val Logs.src_log src : Logs.LOG)
 
-let total_pages = OS.MM.Heap_pages.total ()
-let pagesize_kb = Io_page.page_size / 1024
+let wordsize_in_bytes = Sys.word_size / 8
 
-let meminfo ~used =
-  let mem_total = total_pages * pagesize_kb in
-  let mem_free = (total_pages - used) * pagesize_kb in
-  Log.info (fun f -> f "Writing meminfo: free %d / %d kB (%.2f %%)"
-    mem_free mem_total (float_of_int mem_free /. float_of_int mem_total *. 100.0));
+let fraction_free stats =
+  let { OS.Memory.free_words; heap_words; _ } = stats in
+  float free_words /. float heap_words
+
+let meminfo stats =
+  let { OS.Memory.free_words; heap_words; _ } = stats in
+  let mem_total = heap_words * wordsize_in_bytes in
+  let mem_free = free_words * wordsize_in_bytes in
+  Log.info (fun f -> f "Writing meminfo: free %a / %a (%.2f %%)"
+    Fmt.bi_byte_size mem_free
+    Fmt.bi_byte_size mem_total
+    (fraction_free stats *. 100.0));
   Printf.sprintf "MemTotal: %d kB\n\
                   MemFree: %d kB\n\
                   Buffers: 0 kB\n\
                   Cached: 0 kB\n\
                   SwapTotal: 0 kB\n\
-                  SwapFree: 0 kB\n" mem_total mem_free
+                  SwapFree: 0 kB\n" (mem_total / 1024) (mem_free / 1024)
 
-let report_mem_usage used =
+let report_mem_usage stats =
   Lwt.async (fun () ->
     let open OS in
     Xs.make () >>= fun xs ->
     Xs.immediate xs (fun h ->
-      Xs.write h "memory/meminfo" (meminfo ~used)
+      Xs.write h "memory/meminfo" (meminfo stats)
     )
   )
 
 let init () =
   Gc.full_major ();
-  let used = OS.MM.Heap_pages.used () in
-  report_mem_usage used
+  let stats = OS.Memory.quick_stat () in
+  report_mem_usage stats
 
 let status () =
-  let used = OS.MM.Heap_pages.used () |> float_of_int in
-  let frac = used /. float_of_int total_pages in
-  if frac < 0.9 then `Ok
+  let stats = OS.Memory.quick_stat () in
+  if fraction_free stats > 0.1 then `Ok
   else (
     Gc.full_major ();
-    let used = OS.MM.Heap_pages.used () in
-    report_mem_usage used;
-    let frac = float_of_int used /. float_of_int total_pages in
-    if frac > 0.9 then `Memory_critical
+    let stats = OS.Memory.quick_stat () in
+    report_mem_usage stats;
+    if fraction_free stats < 0.1 then `Memory_critical
     else `Ok
   )
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 # Pin the base image to a specific hash for maximum reproducibility.
-# It will probably still work on newer images, though, unless Debian
+# It will probably still work on newer images, though, unless an update
 # changes some compiler optimisations (unlikely).
-#FROM ocurrent/opam:alpine-3.10-ocaml-4.10
-FROM ocurrent/opam@sha256:4546b41a99b54f163af435327c86f88d06346f2a059f0f42bea431b37329ea8d
+#FROM ocurrent/opam:fedora-32-ocaml-4.10
+FROM ocurrent/opam@sha256:2e0e1689d2260c202bf944034f15ba8ebe945dba6b126cc6dd6b185c223014f3
 
 # Pin last known-good version for reproducible builds.
 # Remove this line (and the base image pin above) if you want to test with the
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,14 +2,14 @@
 # It will probably still work on newer images, though, unless Debian
 # changes some compiler optimisations (unlikely).
 #FROM ocurrent/opam:alpine-3.10-ocaml-4.10
-FROM ocurrent/opam@sha256:d30098ff92b5ee10cf7c11c17f2351705e5226a6b05aa8b9b7280b3d87af9cde
+FROM ocurrent/opam@sha256:4546b41a99b54f163af435327c86f88d06346f2a059f0f42bea431b37329ea8d
 
 # Pin last known-good version for reproducible builds.
 # Remove this line (and the base image pin above) if you want to test with the
 # latest versions.
-RUN cd ~/opam-repository && git fetch origin master && git reset --hard e81ab2996896b21cba74c43a903b305a5a6341ef && opam update
+RUN cd ~/opam-repository && git fetch origin master && git reset --hard 6ef290f5681b7ece5d9c085bcf0c55268c118292 && opam update
 
-RUN opam depext -i -y mirage.3.8.0 lwt.5.3.0
+RUN opam depext -i -y mirage
 RUN mkdir /home/opam/qubes-mirage-firewall
 ADD config.ml /home/opam/qubes-mirage-firewall/config.ml
 WORKDIR /home/opam/qubes-mirage-firewall
--- a/build-with-docker.sh
+++ b/build-with-docker.sh
@@ -5,5 +5,5 @@ docker build -t qubes-mirage-firewall .
 echo Building Firewall...
 docker run --rm -i -v `pwd`:/home/opam/qubes-mirage-firewall qubes-mirage-firewall
 echo "SHA2 of build:   $(sha256sum qubes_firewall.xen)"
-echo "SHA2 last known: 0f6b41fa3995afccff1809cb893c45c0863477d4dfacc441c11e3382bec31d39"
+echo "SHA2 last known: a635ead410ffb72abb8b44e8c5f8f2cfc8752c4787e737ed6cdc0089143ace00"
 echo "(hashes should match for released versions)"
--- a/build-with-docker.sh
+++ b/build-with-docker.sh
@@ -5,5 +5,5 @@ docker build -t qubes-mirage-firewall .
 echo Building Firewall...
 docker run --rm -i -v `pwd`:/home/opam/qubes-mirage-firewall qubes-mirage-firewall
 echo "SHA2 of build:   $(sha256sum qubes_firewall.xen)"
-echo "SHA2 last known: a635ead410ffb72abb8b44e8c5f8f2cfc8752c4787e737ed6cdc0089143ace00"
+echo "SHA2 last known: 583d22327500fa092f436af1d0d9b1b78ebe12abd814c128ec7452c2f4cf319a"
 echo "(hashes should match for released versions)"
--- a/config.ml
+++ b/config.ml
@@ -33,7 +33,7 @@ let main =
       package "mirage-qubes" ~min:"0.8.2";
       package "mirage-nat" ~min:"2.2.1";
       package "mirage-logs";
-      package "mirage-xen" ~min:"5.0.0";
+      package "mirage-xen" ~min:"6.0.0";
       package ~min:"4.5.0" "dns-client";
       package "pf-qubes";
     ]


mato:
> #0  mirage_xen_evtchn_block_domain (v_deadline=<optimized out>) at bindings/evtchn.c:343

That's err, odd. Looks like it's just waiting for an event, but never gets one. Event delivery is very much working for me, tested also with mirage-skeleton device-usage/network, and presumably also for @palainp as we can see in mirage/mirage#1184.

You are on the latest Qubes OS 4, right?

mato:
> You are on the latest Qubes OS 4, right?

...specifically, what is your Xen version, as reported by `xl info` in Dom0?

talex5:
This is Qubes 4.0, `xen_version : 4.8.5-19.fc25`. I've enabled the `qubes-dom0-current-testing` repo too.

mato:
Ok, so that should be fine. Given that you have the version with `type=pvh` support fully added in, could you try and run `device-usage/network` from mirage-skeleton? You'll need to create a bridge in dom0 for it to connect to, but otherwise it should just work. If it works (and test it with e.g. `ping -f`) then at least we can rule out low-level event channels being broken.

talex5:
I tried pinging it from sys-net (that seemed easier). It worked for a bit, then stopped:
```
[user@sys-net ~]$ sudo ping -f 10.137.0.11 
PING 10.137.0.11 (10.137.0.11) 56(84) bytes of data.
........................................................................................................................................................................................................................................................................................^C
--- 10.137.0.11 ping statistics ---
1521 packets transmitted, 1241 received, 18.4089% packet loss, time 13298ms
```

Sometimes it doesn't complete booting. e.g. here are two runs where the first one failed:

```
~/w/m/s/d/network (master) $ test-mirage network.xen mirage-test
Waiting for 'Ready'... OK
Uploading 'network.xen' (6207232 bytes) to "mirage-test"
Waiting for 'Booting'... OK
Connecting to mirage-test console...
Solo5: Xen console: port 0x2, ring @0x00000000FEFFF000
Solo5: using E820 memory map
            |      ___|
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
Solo5: Bindings version v0.6.6-2-g20196872
Solo5: Memory map: 32 MB addressable:
Solo5:   reserved @ (0x0 - 0xfffff)
Solo5:       text @ (0x100000 - 0x293fff)
Solo5:     rodata @ (0x294000 - 0x2e5fff)
Solo5:       data @ (0x2e6000 - 0x3fffff)
Solo5:       heap >= 0x400000 < stack < 0x2000000
2020-08-19 15:02:12 -00:00: INF [net-xen frontend] connect 0
2020-08-19 15:02:12 -00:00: INF [net-xen frontend] create: id=0 domid=1
2020-08-19 15:02:12 -00:00: INF [net-xen frontend]  sg:true gso_tcpv4:true rx_copy:true rx_flip:false smart_poll:false
2020-08-19 15:02:12 -00:00: INF [net-xen frontend] MAC: 00:16:3e:5e:6c:00
^C

~/w/m/s/d/network (master) $ test-mirage network.xen mirage-test
Waiting for 'Ready'... OK
Uploading 'network.xen' (6207232 bytes) to "mirage-test"
Waiting for 'Booting'... OK
Connecting to mirage-test console...
Solo5: Xen console: port 0x2, ring @0x00000000FEFFF000
Solo5: using E820 memory map
            |      ___|
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
Solo5: Bindings version v0.6.6-2-g20196872
Solo5: Memory map: 32 MB addressable:
Solo5:   reserved @ (0x0 - 0xfffff)
Solo5:       text @ (0x100000 - 0x293fff)
Solo5:     rodata @ (0x294000 - 0x2e5fff)
Solo5:       data @ (0x2e6000 - 0x3fffff)
Solo5:       heap >= 0x400000 < stack < 0x2000000
2020-08-19 15:03:31 -00:00: INF [net-xen frontend] connect 0
2020-08-19 15:03:31 -00:00: INF [net-xen frontend] create: id=0 domid=1
2020-08-19 15:03:31 -00:00: INF [net-xen frontend]  sg:true gso_tcpv4:true rx_copy:true rx_flip:false smart_poll:false
2020-08-19 15:03:31 -00:00: INF [net-xen frontend] MAC: 00:16:3e:5e:6c:00
2020-08-19 15:03:31 -00:00: INF [ethernet] Connected Ethernet interface 00:16:3e:5e:6c:00
2020-08-19 15:03:31 -00:00: INF [ARP] Sending gratuitous ARP for 10.137.0.11 (00:16:3e:5e:6c:00)
2020-08-19 15:03:31 -00:00: INF [udp] UDP interface connected on 10.137.0.11
2020-08-19 15:03:31 -00:00: INF [tcpip-stack-direct] stack assembled: mac=00:16:3e:5e:6c:00,ip=10.137.0.11
```

palainp:
Not sure what I've done wrong, but I got : 
```
Solo5: Memory map: 64 MB addressable:
Solo5:   reserved @ (0x0 - 0xfffff)
Solo5:       text @ (0x100000 - 0x2affff)
Solo5:     rodata @ (0x2b0000 - 0x308fff)
Solo5:       data @ (0x309000 - 0x43afff)
Solo5:       heap >= 0x43b000 < stack < 0x4000000
network: too many arguments, don't know what to do with `None'
Usage: network [OPTION]... 
Try `network --help' for more information.
Solo5: solo5_exit(64) called
```

talex5:
@palainp : I got that too. Somehow, the default Qubes kernel arguments were the literal string "None", rather than the empty string. Can be fixed with `qvm-prefs app kernelopts ''`.

palainp:
@talex5 thank you, I was pretty sure that I had allready run this days ago.
With that fixed I can confirm that the vm runs normally (each times) and I can ping it from sys-firewall. I have ```xen_version            : 4.8.5-22.fc25```.

palainp:
I do not use docker, but it seemes to me that the gcc canaries are a a good thing have ?
Surprinsingly, the crash does not occurs when I do not hide the qrexec symbol (understand : use ```qrexec >>= fun qrexec2 ->``` instead of ```qrexec >>= fun qrexec ->```). It's the same for qubes-mirage-firewall, the vm starts correctly, respond to pings, but I still have the ```Netif: ERROR response``` when trying to run wget commands.

palainp:
Some progress : I tracked down the code to a vchan_xen disconnection which may interfere with the unikernel. With the diff below for mirage-qubes I do not have the crash anymore and I have the following log using ```qvm-prefs mirage-test kernelopts '-l "*:debug"'```. Note that it may introduce a memory leak as I am not sure when the flow must be created and destroyed (@talex5 the two send calls in the first block seems to resolve correctly, have you an idea about how to track the bug further ?).
With this patch on, qubes-mirage-firwall unikernel boots too but it still crash with ```Netif: ERROR response``` when I try to download anything or when I try to reconnect the unikernel vm after a crash (with a frameQ decr from 10 to 0).
```
[user@my-new-qube mirage-qubes]$ git diff
diff --git a/lib/rExec.ml b/lib/rExec.ml
index 67e4ffc..32cc411 100644
--- a/lib/rExec.ml
+++ b/lib/rExec.ml
@@ -117,7 +117,8 @@ module Flow = struct
         send flow.dstream ~ty:`Data_exit_code msg >>!= fun () ->
         return (`Ok ())
       )
-      (fun () -> QV.disconnect flow.dstream)
+      (Log.debug(fun f -> f "We were supposed to close the flow here"); fun () -> return (()))
+      (*(fun () -> QV.disconnect flow.dstream)*)
 end
 
 module Client_flow = struct
```
```
Solo5: Xen console: port 0x2, ring @0x00000000FEFFF000
Solo5: using E820 memory map
            |      ___|
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
Solo5: Bindings version v0.6.6-2-g20196872
Solo5: Memory map: 64 MB addressable:
Solo5:   reserved @ (0x0 - 0xfffff)
Solo5:       text @ (0x100000 - 0x2affff)
Solo5:     rodata @ (0x2b0000 - 0x308fff)
Solo5:       data @ (0x309000 - 0x43afff)
Solo5:       heap >= 0x43b000 < stack < 0x4000000
2020-08-26 13:13:43 -00:00: INF [net-xen frontend] connect 0
2020-08-26 13:13:43 -00:00: INF [qubes.db] connecting to server...
2020-08-26 13:13:43 -00:00: INF [qubes.db] connected
2020-08-26 13:13:43 -00:00: INF [net-xen frontend] create: id=0 domid=3
2020-08-26 13:13:43 -00:00: INF [net-xen frontend]  sg:true gso_tcpv4:true rx_copy:true rx_flip:false smart_poll:false
2020-08-26 13:13:43 -00:00: INF [net-xen frontend] MAC: 00:16:3e:5e:6c:00
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/type" = "AppVM"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-vm-updateable" = "False"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-vm-type" = "ProxyVM"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-vm-persistence" = "rw-only"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-usb-devices" = ""
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-timezone" = "Europe/Paris"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-service/meminfo-writer" = "1"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-secondary-dns" = "10.139.1.2"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-random-seed" = "cW08j3YNke93wFpShrLS7mJr3+DYA6CrfpQRzuIlCDB4soH1DxoPkk2KRNRGbhlvI7eH1YG2s0VNt0fAEiD9OQ=="
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-primary-dns" = "10.139.1.1"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-netvm-secondary-dns" = "10.139.1.2"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-netvm-primary-dns" = "10.139.1.1"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-netvm-network" = "10.137.0.50"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-netvm-netmask" = "255.255.255.255"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-netvm-gateway" = "10.137.0.50"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-netmask" = "255.255.255.255"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-mac" = "00:16:3e:5e:6c:00"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-iptables-error" = ""
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-ip" = "10.137.0.50"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-gateway" = "10.137.0.3"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-debug-mode" = "1"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-block-devices" = ""
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/qubes-base-template" = "fedora-32"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/name" = "mirage-test"
2020-08-26 13:13:43 -00:00: DBG [qubes.db] "/default-user" = "user"
2020-08-26 13:13:43 -00:00: INF [ethernet] Connected Ethernet interface 00:16:3e:5e:6c:00
2020-08-26 13:13:43 -00:00: INF [ARP] Sending gratuitous ARP for 10.137.0.50 (00:16:3e:5e:6c:00)
2020-08-26 13:13:43 -00:00: INF [udp] UDP interface connected on 10.137.0.50
2020-08-26 13:13:43 -00:00: INF [tcpip-stack-direct] stack assembled: mac=00:16:3e:5e:6c:00,ip=10.137.0.50
2020-08-26 13:13:43 -00:00: DBG [tcpip-stack-direct] Establishing or updating listener for stack mac=00:16:3e:5e:6c:00,ip=10.137.0.50
2020-08-26 13:13:43 -00:00: INF [qubes.rexec] waiting for client...
2020-08-26 13:13:43 -00:00: INF [qubes.rexec] client connected, using protocol version 2
2020-08-26 13:13:43 -00:00: INF [application] ready to listen for new packets
2020-08-26 13:13:43 -00:00: DBG [tcpip-stack-direct] Establishing or updating listener for stack mac=00:16:3e:5e:6c:00,ip=10.137.0.50
2020-08-26 13:13:44 -00:00: DBG [qubes.rexec] Execute "user:QUBESRPC qubes.SetMonitorLayout dom0\000"
2020-08-26 13:13:44 -00:00: INF [application] QUBESRPC qubes.SetMonitorLayout dom0
2020-08-26 13:13:44 -00:00: DBG [qubes.rexec] "QUBESRPC qubes.SetMonitorLayout dom0" returned exit status 0
2020-08-26 13:13:44 -00:00: DBG [qubes.rexec] We were supposed to close the flow here
2020-08-26 13:13:52 -00:00: DBG [qubes.rexec] Execute "user:QUBESRPC qubes.WaitForSession dom0\000"
2020-08-26 13:13:52 -00:00: DBG [qubes.rexec] "QUBESRPC qubes.WaitForSession dom0" returned exit status 0
2020-08-26 13:13:52 -00:00: DBG [qubes.rexec] We were supposed to close the flow here
```

mato:
@talex5 

> I also see the crash when compiled using the native Fedora 32 Qubes compiler, but not when compiling in a Fedora 32 Docker container. I'm using OCaml 4.10.0 in both cases. Opam package versions are the same. The generated code is different. For example, the Qubes one has references to __stack_chk_guard.

Interesting. The Solo5 + ocaml-freestanding build infrastructure should be force-enabling stack canaries on all compilers, so it's odd that the Docker-compiled version does not have any. Having said that, the generated assembly looks fine, and if the generated SSP code were triggering the malloc corruption _directly_, I'd expect to see aborts due to SSP failures as well.

@palainp 

> Some progress : I tracked down the code to a vchan_xen disconnection which may interfere with the unikernel.

Ah, that's useful. @talex5 pointed out to me that the vchan test unikernels do not test any disconnections. It'd be useful to add that. @djs55, feel like some hacking? :-)

talex5:
Ah, looks like I messed up the testing somehow. An image built with the Docker Fedora base does include the stack check, but building with Alpine doesn't. Sorry for the confusion!

talex5:
Commenting out this line in mirage-xen seems to help:
```ocaml
free(mapping->start_addr);
```

With that, the network VM survived a flood ping, and the firewall works.

https://github.com/mato/mirage-xen/blob/528d6c2d83cc49b363b21da8049e0089cd0c173b/lib/bindings/gnttab.c#L320

palainp:
@talex5 great job, I can confirm that it starts correctly and I am able to get some web pages with firefox, but the unikernel crashes when I try to load some big traffic (for example : ```https://www.nperf.com/en/```) with ```qvm-prefs memory``` and ```maxmem``` set to 64MB with this logs (I'm pretty sure that we can get rid of this bug soon ! :stuck_out_tongue:):
```
Io_page: memalign(4096, 4096) failed, even after GC.
Fatal error: exception Out of memory
Raised by primitive operation at file "lib/io_page.ml", line 42, characters 8-27
Called from file "lib/io_page.ml", line 44, characters 4-22
Called from file "lib/io_page.ml", line 59, characters 25-32
Called from file "lib/frontend.ml", line 152, characters 18-35
Called from file "lib/frontend.ml", line 231, characters 6-26
Called from file "src/core/lwt.ml", line 1867, characters 23-26
Re-raised at file "src/core/lwt.ml", line 3027, characters 20-29
Called from file "lib/main.ml", line 36, characters 10-20
Called from file "main.ml" (inlined), line 5, characters 12-25
Called from file "main.ml", line 80, characters 5-10
Solo5: solo5_exit(2) called
```

talex5:
Here's a version that leaks address space rather than RAM (`0x40000000` is the end of physical memory for me, based on the memory map solo5 outputs at boot time; adjust as appropriate):

https://github.com/talex5/mirage-xen/commit/5e4c6b598e68cab90ac6710f7c395f076b04f43a

I was hoping to trigger a page fault when the use-after-free happens, but it seems to be running fine.

palainp:
With this call to free deleted the unikernel still may encounter an memory allocation problem even if it's later now right? (unless the ocmal GC can deals with this?)
However, downloading a large file does not seem to be a problem now, but even if I have removed all the printed informations for limitating the impact of IOs, I observe a drop in speed of about 30% and a load that goes from negligible (for sys-firewall fedora 32) at 99% CPU for an http download and for a load test of my internet connection a division by 300 of the speed with a CPU load going from 25% to 99%. As I think it's in majority due to the CPU load is there any tool like callgrind that can I play with for finding the bottlenecks?

talex5:
> With this call to free deleted the unikernel still may encounter an memory allocation problem even if it's later now right? 

In my branch I removed the alloc call too, so it won't run out of RAM. This memory isn't actually needed for anything; the solo5 mirage-xen branch was just (wastefully) allocating memory as a way a reserving some memory addresses. My test branch just keeps increasing a pointer to get fresh addresses. Eventually it will wrap and Bad Things will happen, but it's a 64-bit address incremented in page-sized chunks, so it might handle about 2^52 allocations before that.

> As I think it's in majority due to the CPU load is there any tool like callgrind that can I play with for finding the bottlenecks?

I'm not sure. There's some (old) information about getting a flamegraph for a Xen Mirage unikernel here:

http://www.brendangregg.com/blog/2016-01-27/unikernel-profiling-from-dom0.html

I also did a few profiling experiments years ago, but I suspect they'd need quite a bit of updating to work now:

https://roscidus.com/blog/blog/2014/08/15/optimising-the-unikernel/

palainp:
Thanks for this very interesting links. I managed to find that the CPU overload is due to a massive communication between the unikernel vm and dom0 (during a test using ```curl -4 -o /dev/null https://mirrors.edge.kernel.org/qubes/iso/Qubes-R4.0.3-x86_64.iso``` I got theese 3 dom0 processes using about 130% CPU : ```qmemmand```, ```xenstored``` and ```systemd-journal``` with a lots of writing in the qubes' logs).
This massive dialog between the unikernel and dom0 is caused by the "divide by 0" in the memory_pressure.ml status function. We have a NaN compared to 0.9 which results in using the else statement that triggers the garbage collector and a memory usage report to xen (and that fills the logs). We can avoid that by testing the result of the mm_stubs function ```OS.MM.Heap_pages.total``` (see https://github.com/palainp/qubes-mirage-firewall/commit/079f7c34731a27a8dc5eeed6c6f1ac7ed27f161c) and we have with similar performances compared  to the linux kernel but this leads me to an abort call in solo5 after downloading something like 240 to 260MB with a unikernel memory fixed to 64MB.
As we give some adress space pieces to the unikernel and we never free the memory in the unmap function, I think we have to tell the GC that there is something to clean from times to times, but unfortunatly if the status function to returns ```Memory_critical``` the unikernel stops running after the ```Memory low - dropping packet and resetting NAT table``` before reseting ```My_nat.reset```.

talex5:
Ah, the NaN causing extra communication due to memory stats reporting makes sense - nice find!

We probably shouldn't be doing a GC manually, but there were some bugs in the past (maybe fixed now; not sure):

- https://github.com/mirage/io-page/issues/38
- https://github.com/ocaml/ocaml/issues/7100

BTW, @mato pointed me at https://github.com/Solo5/solo5/blob/master/bindings/xen/platform.c#L74, which says:

```c
    /*
     * Highest mapped physical address, currently 4GB. Note that Xen maps
     * various structures just below 4GB in the guest-physical address space.
     */
    uint64_t paddr_end = PLATFORM_MAX_MEM_SIZE;
```

So probably we'll fail when we get near to 4 GB, long before exhausting the 64-bit address space. And maybe the page tables beyond that aren't set up either.

I guess solo5 should tell OCaml a range of addresses that it isn't using, and OCaml should use that to work out where to map things and pass the address to the solo5 map function.

mato:
> So probably we'll fail when we get near to 4 GB, long before exhausting the 64-bit address space. And maybe the page tables beyond that aren't set up either.

Yes we will (fail when we get near to 4GB), since we'll hit Xen PV rings / ACPI / PCI memory mapped at around <4GB. And yes, the page tables are not set up beyond 4GB.

So, as far as I can tell, we need two things:

1. A way to get a _virtual_ memory address range we can use to allocate pages for imported grants.
2. A simple page allocator for it / other scheme to manage those pages.

(1) has no APIs in Solo5 at the moment, they may come in the future, but since we control the whole stack I'll see what I can make work for Mirage/Xen internal use only. (2) Should be fairly straightforward, any suggestions?

Regarding the values returned by `OS.Mm` via `mm_stubs.c`, I think we can hook those up to something sensible. We have `mallinfo()` from dlmalloc / ocaml-freestanding, which should give us a figure for "Total allocated space (bytes)" in `->uordblks` and we can get the `heap_size` (again, bytes) from `*si` at `solo5_app_main()` call time.

palainp:
> So, as far as I can tell, we need two things:

As a beginner with solo5 I think you will do it better, but for trying your solution I used a bit of the actual heap for allocation usages with the grants (see https://github.com/palainp/solo5/commit/4d6d363c48ee07810120b4dc33dd734679fe878e ) and changed the mallloc/free calls (in https://github.com/palainp/mirage-xen/commit/e6d877af4250b8f6d636cc469e1397a3d47ed5f9 this allocation algorithm is not optimal neither in complexity O(n) whereas we can use some O(log(n)) solution nor in space as it uses 16kio of memory all the time but is really simple to write/understand and debug with).

The qubes-mirage-firewall unikernel runs now correctly with this pinned repo :
```
[user@dev ~]$ opam list | grep pinned
mirage-xen                6.0.0       pinned to version 6.0.0 at git+https://github.com/palainp/mirage-xen.git#with-solo5-internal-heap
solo5-bindings-xen        0.6.6       pinned to version 0.6.6 at git+https://github.com/palainp/solo5.git#support-xen-4.8
```

BTW when searching for the allocation problem, I noticed that the corruption occurs when we use some pages first with the mirage_xen_gnttab_map* function (malloc call)/unmap (free call), when the same page is given (once freed) to ocaml by the  mirage_alloc_pages function. With a separation of the heap this will not occurs and the problem does not appears but a clever solution should exists.

palainp:
Update : I wrote another allocating algorithm using the Knuth TAOCP 'buddy system' for reference. I then discovered that my patch for mm_stubs.c causes the principal loss in terms of performance (CPU load is going from 30% when both ```stub_heap_get_pages_total``` and ```stub_heap_get_pages_used``` returns 0 to 99% for the same network load test when they returns something else. @mato I've definitively not understood what are the correct modification :disappointed:  ).

Since the last couples of days I use qubes-mirage-firewall (with no special filtering rules) as my main firewall with : 
* solo5-bindings-xen that reserve (256+1 for the allocation algorithm) pages at the beginning of the heap (https://github.com/palainp/solo5/commit/b62eb243243b358521f07e53fc5d7fc6edfeda61),
* a buddy system for managing memory allocation in gnttab.c and 
* fake values returned in mm_stubs.c (https://github.com/palainp/mirage-xen/blob/with-solo5-internal-heap/lib/bindings/gnttab.c and https://github.com/palainp/mirage-xen/blob/with-solo5-internal-heap/lib/bindings/mm_stubs.c)

mato:
@palainp @talex5 

Finally have some time to come back to this. Can you try updating to the latest Xen-PVH OPAM repository (`git+https://github.com/mirage/mirage-dev.git#3.x+xen-pvh-via-solo5`, mainly to get you Mirage 3.8.1), and building with the following additional pins:

```
mirage-xen.6.0.0            git  git+https://github.com/mato/mirage-xen#wip-xen-gntmap
solo5-bindings-xen.0.6.6    git  git+https://github.com/mato/solo5#support-xen-4.8
```

There are several work in progress changes in those trees:

1. On the Solo5 bindings side, we provide page tables for an extra 1GB of memory beyond PLATFORM_MAX_MEM_SIZE; this area starting at 4GB can then be used for imported grant mappings (https://github.com/mato/solo5/commit/13585420c27f230a838284ddb83fb3eedd6ab752).
2. On the Mirage/Xen side, I have added a simple bitmap allocator and changed the gntmap functions to use it (https://github.com/mato/mirage-xen/commit/7698d7cdde1b279057ddf0cb765640494955b7d2). I'm not particularly proud of it, but it should be worst-case O(n) for allocations where n is the total amount of pages given to it / 8. I've tested the algorithms semi-manually for edge cases and they seem fine, but would appreciate some review.
3. On the Mirage/Xen side, I've also hooked up `OS.Mm` (https://github.com/mato/mirage-xen/commit/7cea1f30ac8012273e4f923285fea0f022f93a5c) the way I think it should be done for now. (/cc @hannesm, we had a related discussion about this; this is what I meant, though we should probably use bytes rather than pages on the OCaml side).

All comments / review appreciated, if you can confirm that these changes get you a stable qubes-mirage-firewall then I'll work toward cleaning them up and PR'ing appropriately.

palainp:
I will try your changes later this week.

For your second point, I found that the memory requests in gnttab.c is exclusively for 1 or 16 pages (after a bunch of 1 page allocations we have a pattern like this : alloc a=1 page, alloc b=16 pages, alloc c=16 pages, free c, free b, free a. -> It may be possible to have 2 differents areas if we assume to have requests only for 1 or 16 pages).

I tried a buddy allocator that works most of the time in O(1) and in the worst case at O(log(n)) with n the size of the available memory (during my own tests I used the first page for the management of the linked list). The complexity improvement is done with the usage of linked lists of free memory areas (we just have to pop off the first node in the list to get free memory). However I did not notice a significant improvement over a bitmap search method (I think we benefit from the allocation pattern described before and the unikernel almost constantly use the same pages).

palainp:
@mato 

Updates done, I only removed the DPRINTFs in gnttab.c and compiled.

With correct values returned to ocaml by ```mirage_xen_heap_get_pages_total``` and ```mirage_xen_heap_get_pages_used``` I have a 100% CPU load  when trying ```curl -4 -o /dev/null https://mirrors.edge.kernel.org/qubes/iso/Qubes-R4.0.3-x86_64.iso``` whereas with fake values (the stub version that returns 0) and the modification of qubes-mirage-firewall to avoid extra communication due to memory stats reporting, the CPU drop to something like 10-30%.

Apart this point the resulting unikernel seems stable, with my version from https://github.com/mirage/qubes-mirage-firewall/pull/116#issuecomment-693301406 I have experienced some "out of memory" after long use but now, with 1GB for the grants I think it will be fine, I will do more tests and extensive use.

Diff in mirage-xen :
```
--- a/lib/bindings/main.c
+++ b/lib/bindings/main.c
@@ -111,16 +111,18 @@ mirage_xen_get_xenstore_page(value v_unit)
 CAMLprim value
 mirage_xen_heap_get_pages_total(value v_unit)
 {
-    return Val_long(solo5_heap_size / PAGE_SIZE);
+    return Val_long(0);
 }
 
 /* @@noalloc */
 CAMLprim value
 mirage_xen_heap_get_pages_used(value v_unit)
 {
-    struct mallinfo mi = mallinfo();
-    return Val_long(mi.uordblks / PAGE_SIZE);
+    return Val_long(0);
 }
```
and in qubes-mirage-firewall :
```
--- a/memory_pressure.ml
+++ b/memory_pressure.ml
@@ -38,7 +38,7 @@ let init () =
 let status () =
   let used = OS.MM.Heap_pages.used () |> float_of_int in
   let frac = used /. float_of_int total_pages in
-  if frac < 0.9 then `Ok
+  if total_pages==0 || frac < 0.9 then `Ok
   else (
     Gc.full_major ();
     let used = OS.MM.Heap_pages.used () in
```
I currently don't know how to investigate more on this CPU load :disappointed: Could the garbage collector be somehow involved in this overusage?

mato:
@palainp:

Thanks for testing!

> Updates done, I only removed the DPRINTFs in gnttab.c and compiled.

I left those in deliberately to verify that the correct address range was being used.

> With correct values returned to ocaml by mirage_xen_heap_get_pages_total and mirage_xen_heap_get_pages_used I have a 100% CPU load when trying curl -4 -o /dev/null https://mirrors.edge.kernel.org/qubes/iso/Qubes-R4.0.3-x86_64.iso whereas with fake values (the stub version that returns 0) and the modification of qubes-mirage-firewall to avoid extra communication due to memory stats reporting, the CPU drop to something like 10-30%.

I don't know what's up with that. @talex5, any ideas?

> with my version from #116 (comment) I have experienced some "out of memory" after long use but now, with 1GB for the grants I think it will be fine, I will do more tests and extensive use.

It's actually 4k pages, so 16MB, but it should be fine given that there's only 2k entries in the grant table itself. (See https://github.com/mato/mirage-xen/blob/wip-xen-gntmap/lib/bindings/gnttab.c#L56).

> For your second point, I found that the memory requests in gnttab.c is exclusively for 1 or 16 pages (after a bunch of 1 page allocations we have a pattern like this : alloc a=1 page, alloc b=16 pages, alloc c=16 pages, free c, free b, free a. -> It may be possible to have 2 differents areas if we assume to have requests only for 1 or 16 pages).

That happens to be what the qubes-mirage-firewall does, but the `OS.Xen.Import` is a general API, so we can't rely on that.

talex5:
> With correct values returned to ocaml by mirage_xen_heap_get_pages_total and mirage_xen_heap_get_pages_used I have a 100% CPU load

What does the log show? e.g. mine (PVH) says:

```
INF [memory_pressure] Writing meminfo: free 97756 / 126500 kB (77.28 %)
```

It always logs that message when it does a GC, so it should be obvious if the values don't make sense or if it's doing a lot of GC work. If the free memory is less than 10% then it will do a GC.

palainp:
When the kernel starts I have this log (it's pretty close to the 10% limit)
```
INF [memory_pressure] Writing meminfo: free 54268 / 60712 kB (89.39 %)
```
But after that I have no sign of memory_pressure in the logs. I'm not sure that this issue is related with the memory_pressure code.

What I tried is : 
* with the correct total_page and used_page values I have 100% and a smaller bandwidth than the previous test (this is the situation with the up to date mirage-xen code and DPRINTF removed)
```
$ curl -4 -m 60 -o /dev/null https://mirrors.edge.kernel.org/qubes/iso/Qubes-R4.0.3-x86_64.iso
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  7 4607M    7  353M    0     0  6040k      0  0:13:00  0:00:59  0:12:01 7682k
curl: (28) Operation timed out after 60000 milliseconds with 371129958 out of 4830789632 bytes received
```
* with total_page=0 and a test like ```if total_pages==0 || frac < 0.9 then `Ok``` I have about 30% of CPU load in the qubes tray
```
$ curl -4 -m 60 -o /dev/null https://mirrors.edge.kernel.org/qubes/iso/Qubes-R4.0.3-x86_64.iso
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
 19 4607M   19  877M    0     0  14.6M      0  0:05:15  0:00:59  0:04:16 21.8M
curl: (28) Operation timed out after 60000 milliseconds with 919895654 out of 4830789632 bytes received
```
* with the correct total_page and used_page values and the status() in memory_pressure which returns allway Ok (```let status () = `Ok ```) I still have 100% CPU load (this is the test that leads me to the conclusion that it is not related to the GC call in memory_pressure, is this a correct conclusion ?)
```
$ curl -4 -m 60 -o /dev/null https://mirrors.edge.kernel.org/qubes/iso/Qubes-R4.0.3-x86_64.iso
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  7 4607M    7  347M    0     0  5933k      0  0:13:15  0:00:59  0:12:16 7389k
curl: (28) Operation timed out after 60000 milliseconds with 364559974 out of 4830789632 bytes received
```

talex5:
> When the kernel starts I have this log (it's pretty close to the 10% limit)
> INF [memory_pressure] Writing meminfo: free 54268 / 60712 kB (89.39 %)

That's free memory, not used :-)

I wonder if the new `OS.MM.Heap_pages.total` is slow (we do it on every packet). In Mini-OS it's just reading a variable, but maybe the new code has to scan the heap or something? But if I'm understanding correctly, your third test wasn't calling it at all, right? And it was still slow.

mato:
> I wonder if the new OS.MM.Heap_pages.total is slow (we do it on every packet).

According to the dlmalloc documentation (see `nolibc/dlmalloc.i` in ocaml-freestanding), `mallinfo()` can be somewhat expensive to call. There are two other options:

1. dlmalloc provides a `malloc_footprint()` function which is documented as:

      Returns the number of bytes obtained from the system.  The total
  number of bytes allocated by malloc, realloc etc., is less than this
  value. Unlike mallinfo, this function returns only a precomputed
  result, so can be called frequently to monitor memory consumption. 
  Even if locks are otherwise defined, this function does not use them,
  so results might not be up to date.

2. If that turns out to be lacking, the Mirage/Xen bindings could call `sbrk(0)` to determine the current top of the heap, thus computing the total allocated bytes using the known start-of-heap address. This call is cheap.

In both cases the values may not be 100% up to date, in that `malloc()` may have free blocks internally available on its heap which are not accounted for, but there's not much we can do about that.

I'm going to clean up the current commits and branches a bit, and integrate them into the proper PRs, since apart from the heap statistics everything else is ready for review and seems to work. Once I've done that it'll be easier to test the various heap statistics gathering methods separately from Qubes/Xen 4.8 support and the gnttap virtual memory area changes.

palainp:
@talex5 you are absolutely right, the ```let status() = `Ok``` was an oversimplification and I let, by mistake, a call to the ```OS.MM.Heap_pages.used ()``` which is the cause for the CPU load as @mato indicates. I just did a test with and without and the result of the third test becomes the same as the second.

mato:
@palainp I've force-pushed the `wip-xen-gntmap` branch on `mato/mirage-xen` to remove the DPRINTFs and use `malloc_footprint()`. Can you test with that (and with an unmodified qubes-mirage-firewall)? Also check that the logs from the memory_pressure thread make sense and match what you'd expect in terms of used/free memory?

palainp:
@mato I think it's good now (i.e. no more 100% CPU load). The free (thanks @talex5 :)) memory reported is a bit lower than previously as expected.

Here is my opam configuration : 
```
$ opam repo
[NOTE] These are the repositories in use by the current switch. Use '--all' to
       see all configured repositories.

<><> Repository configuration for switch default ><><><><><><><><><><><><><><><>
 1 mirage-dev-3.x+xen-pvh-via-solo5 git+https://github.com/mirage/mirage-dev.git#3.x+xen-pvh-via-solo5
 2 default                          https://opam.ocaml.org
$ opam list | grep pinned
mirage-xen                6.0.0       pinned to version 6.0.0 at git+https://github.com/mato/mirage-xen#wip-xen-gntmap
solo5-bindings-xen        0.6.6       pinned to version 0.6.6 at git+https://github.com/mato/solo5#support-xen-4.8
```
I have the following logs : 
```
Solo5: Xen console: port 0x2, ring @0x00000000FEFFF000
Solo5: using E820 memory map
            |      ___|
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
Solo5: Bindings version v0.6.6-3-g13585420
Solo5: Memory map: 64 MB addressable:
Solo5:   reserved @ (0x0 - 0xfffff)
Solo5:       text @ (0x100000 - 0x2ebfff)
Solo5:     rodata @ (0x2ec000 - 0x34efff)
Solo5:       data @ (0x34f000 - 0x4b5fff)
Solo5:       heap >= 0x4b6000 < stack < 0x4000000
2020-10-07 10:34:32 -00:00: INF [qubes.rexec] waiting for client...
2020-10-07 10:34:32 -00:00: INF [qubes.gui] waiting for client...
2020-10-07 10:34:32 -00:00: INF [qubes.db] connecting to server...
2020-10-07 10:34:32 -00:00: INF [qubes.db] connected
2020-10-07 10:34:32 -00:00: INF [qubes.rexec] client connected, using protocol version 2
2020-10-07 10:34:32 -00:00: INF [unikernel] QubesDB and qrexec agents connected in 0.033 s
2020-10-07 10:34:32 -00:00: INF [dao] Got network configuration from QubesDB:
            NetVM IP on uplink network: 10.137.0.3
            Our IP on uplink network:   10.137.0.50
            Our IP on client networks:  10.137.0.50
2020-10-07 10:34:32 -00:00: INF [net-xen frontend] connect 0
2020-10-07 10:34:32 -00:00: INF [net-xen frontend] create: id=0 domid=3
2020-10-07 10:34:32 -00:00: INF [net-xen frontend]  sg:true gso_tcpv4:true rx_copy:true rx_flip:false smart_poll:false
2020-10-07 10:34:32 -00:00: INF [net-xen frontend] MAC: 00:16:3e:5e:6c:00
2020-10-07 10:34:32 -00:00: INF [ethernet] Connected Ethernet interface 00:16:3e:5e:6c:00
2020-10-07 10:34:32 -00:00: INF [ARP] Sending gratuitous ARP for 10.137.0.50 (00:16:3e:5e:6c:00)
2020-10-07 10:34:32 -00:00: INF [ARP] Sending gratuitous ARP for 10.137.0.50 (00:16:3e:5e:6c:00)
2020-10-07 10:34:32 -00:00: INF [udp] UDP interface connected on 10.137.0.50
2020-10-07 10:34:32 -00:00: INF [dao] Watching backend/vif
2020-10-07 10:34:32 -00:00: INF [memory_pressure] Writing meminfo: free 53204 / 60712 kB (87.63 %)
2020-10-07 10:34:32 -00:00: INF [qubes.db] got update: "/qubes-keyboard" = "xkb_keymap {\n\txkb_keycodes  { include \"evdev+aliases(azerty)\"\t};\n\txkb_types     { include \"complete\"\t};\n\txkb_compat    { include \"complete\"\t};\n\txkb_symbols   { include \"pc+fr(oss)+inet(evdev)\"\t};\n\txkb_geometry  { include \"pc(pc105)\"\t};\n};"
2020-10-07 10:34:32 -00:00: INF [qubes.gui] client connected (screen size: 1368x768 depth: 24 mem: 4105x)
2020-10-07 10:34:32 -00:00: INF [unikernel] GUI agent connected
2020-10-07 10:34:32 -00:00: WRN [command] << Unknown command "QUBESRPC qubes.SetMonitorLayout dom0"
2020-10-07 10:34:57 -00:00: INF [client_net] add client vif {domid=8;device_id=0} with IP 10.137.0.14
2020-10-07 10:34:57 -00:00: INF [qubes.db] got update: "/mapped-ip/10.137.0.14/visible-ip" = "10.137.0.14"
2020-10-07 10:34:57 -00:00: INF [qubes.db] got update: "/mapped-ip/10.137.0.14/visible-gateway" = "10.137.0.50"
2020-10-07 10:34:57 -00:00: INF [qubes.db] got update: "/qubes-firewall/10.137.0.14/policy" = "drop"
2020-10-07 10:34:57 -00:00: INF [qubes.db] got update: "/qubes-firewall/10.137.0.14/0000" = "action=accept"
2020-10-07 10:34:57 -00:00: INF [qubes.db] got update: "/qubes-firewall/10.137.0.14" = ""
2020-10-07 10:35:01 -00:00: INF [client_net] Client 8 (IP: 10.137.0.14) ready
2020-10-07 10:35:01 -00:00: INF [ethernet] Connected Ethernet interface fe:ff:ff:ff:ff:ff
2020-10-07 10:35:04 -00:00: INF [qubes.db] got rm "/qubes-firewall/10.137.0.14/"
2020-10-07 10:35:04 -00:00: INF [qubes.db] got update: "/qubes-firewall/10.137.0.14/policy" = "drop"
2020-10-07 10:35:04 -00:00: INF [qubes.db] got update: "/qubes-firewall/10.137.0.14/0000" = "action=accept"
2020-10-07 10:35:04 -00:00: INF [qubes.db] got update: "/qubes-firewall/10.137.0.14" = ""
2020-10-07 10:35:23 -00:00: INF [client_eth:dom8:10.137.0.14] who-has 10.137.0.50? responding with fe:ff:ff:ff:ff:ff
2020-10-07 10:36:16 -00:00: INF [client_eth:dom8:10.137.0.14] who-has 10.137.0.50? responding with fe:ff:ff:ff:ff:ff
```
and the curl call which gives me the best results : 
```
$ curl -4 -m 60 -o /dev/null https://mirrors.edge.kernel.org/qubes/iso/Qubes-R4.0.3-x86_64.iso
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
 19 4607M   19  875M    0     0  14.5M      0  0:05:15  0:00:59  0:04:16 18.6M
curl: (28) Operation timed out after 60000 milliseconds with 918519398 out of 4830789632 bytes received
```

mato:
Okay, I've cleaned up the various trees and updated the PVH OPAM repository in mirage/mirage#1159 to match.

@palainp This means that you should be able to rebuild against that repository without any pins and get the same working versions as what you've been testing with thus far, plus some cosmetic improvements I did along the way. Please confirm that works as expected.

palainp:
@mato yes it works fine as expected with just the repository ```mirage-dev-3.x+xen-pvh-via-solo5```. 
I'm starting to use it as a daily firewall VM to continue testing.

palainp:
The kernel works as expected, but after a few hours of running (not big network usage, just an active page on gmail.com or an opened app like discord) it gets "out of memory" messages that makes it impossible to use. I'll try to investigate a bit to find out if it's the estimation by ```malloc_footprint()``` that's too pessimistic or something else. Maybe 64MB is not enough too.
```
2020-10-15 06:33:06 -00:00: INF [memory_pressure] Writing meminfo: free 6068 / 60712 kB (9.99 %)
2020-10-15 06:33:06 -00:00: WRN [firewall] Memory low - dropping packet and resetting NAT table
2020-10-15 06:33:06 -00:00: INF [memory_pressure] Writing meminfo: free 6068 / 60712 kB (9.99 %)
2020-10-15 06:33:06 -00:00: WRN [firewall] Memory low - dropping packet and resetting NAT table
2020-10-15 06:33:08 -00:00: INF [memory_pressure] Writing meminfo: free 6068 / 60712 kB (9.99 %)
2020-10-15 06:33:08 -00:00: WRN [firewall] Memory low - dropping packet and resetting NAT table
2020-10-15 06:33:08 -00:00: INF [memory_pressure] Writing meminfo: free 6068 / 60712 kB (9.99 %)
2020-10-15 06:33:08 -00:00: WRN [firewall] Memory low - dropping packet and resetting NAT table
```

hannesm:
@palainp thanks for your report -- I guess you're using mirage 3.8.1 (and OCaml >= 4.10) -- would you mind to try passing `--allocation-policy=best-fit` to the unikernel (which then uses the best fit allocator for OCaml data, which in my experiments works much nicer).

On a related topic, there's a statistical memory profiler in OCaml (but not yet supported with MirageOS) -- [this blog article](https://blog.janestreet.com/finding-memory-leaks-with-memtrace/) has a nice description on how to use it for finding memory leaks. I'd soon like to play a bit with load tests of qubes-mirage firewall and see whether I'm able to spot memory leaks. Since you mention "long running, not many VMs start & stop", do you have an intuition whether there are lots of network packets (and/or network flows) involved?

palainp:
I'm still not sure what's causing the low memory status (if it's related to a leak, a high number of concurrent flows or a too pesimistic estimation of the free memory) and it's hard to set it up.

For now I'm trying the kernel with this patch to qubes-mirage-firewall (plus the ```really_used()``` function in mirage-xen using ```mallinfo()``` as in https://github.com/mato/mirage-xen/commit/7cea1f30ac8012273e4f923285fea0f022f93a5c) to check against the third option : 
```
[user@dev qubes-mirage-firewall]$ git diff
diff --git a/memory_pressure.ml b/memory_pressure.ml
index ed5b7e5..5125265 100644
--- a/memory_pressure.ml
+++ b/memory_pressure.ml
@@ -42,8 +42,10 @@ let status () =
   else (
     Gc.full_major ();
     let used = OS.MM.Heap_pages.used () in
-    report_mem_usage used;
-    let frac = float_of_int used /. float_of_int total_pages in
+    let really_used = OS.MM.Heap_pages.really_used () in
+    report_mem_usage really_used;
+    Log.info (fun f -> f "memory reported: used %d / really_used %d kB" used really_used);
+    let frac = float_of_int really_used /. float_of_int total_pages in
     if frac > 0.9 then `Memory_critical
     else `Ok
   )
```

mato:
@palainp FYI, mirage 3.9.0 is now released to OPAM with all the Xen changes. This PR will need to be updated for the OS.MM interface changes (mirage/mirage-xen#26) but otherwise you should be able to run the firewall now with no pins or custom OPAM repositories.

memory_pressure.ml
```diff
@@ -6,44 +6,46 @@ open Lwt
...
-  let mem_free = (total_pages - used) * pagesize_kb in
+let fraction_free stats =
+  let { OS.Memory.free_words; heap_words; _ } = stats in
+  float free_words /. float heap_words
+
+let meminfo stats =
+  let { OS.Memory.free_words; heap_words; _ } = stats in
+  let mem_total = heap_words * wordsize_in_bytes in
+  let mem_free = free_words * wordsize_in_bytes in
   Log.info (fun f -> f "Writing meminfo: free %d / %d kB (%.2f %%)"
```
hannesm, line 19:
```suggestion
  Log.info (fun f -> f "Writing meminfo: free %d / %d B (%.2f %%)"
```

previously, pages were reported and kilobytes printed, now words are reported and bytes printed. It may on the other side make sense to still report kilobytes (i.e. divide by 1024 above).

talex5: Good point. I've updated it to use `Fmt.bi_byte_size` instead (though still untested, because it doesn't finish booting for me).

talex5:
I've switched the base image in the Dockerfile from Alpine to Fedora, and it seems to be working now.

hannesm (approved):

README.md
```diff
@@ -75,16 +76,27 @@ qvm-create \
...
+**Note**: for `virt_mode`, use `pv` instead of `pvh` for firewall versions before 0.8.
+
+## Upgrading
+
 To upgrade from an earlier release, just overwrite `/var/lib/qubes/vm-kernels/mirage-firewall/vmlinuz` with the new version and restart the firewall VM.
 
+If upgrading from a version before 0.8, you will also need to switch `virt_mode` from `pv` to `pvh`:
+
+```
+qvm-prefs mirage-firewall virt_mode pvh
```
hannesm, line 97: eventually we should advertise to set
`qvm-features mirage-firewall no-default-kernelopts 1`
and setting the kernelopts to the empty string (used to be `None` in earlier instructions)?

talex5: Good idea - done.

xaki23:
build of 03c7cba (so i was missing the last readme update) through qubes-builder(*) _just_ _worked_ ... 
resulting kernel deployed as dom0-kernel, fired up a standalone based on it, _just_ _works_ ...
now deployed in the usual "lets try this out" minor role (dedicated mfw for my "goes up and down every hour" external-backup-sync vm), first push seems to _just_ _work_ ...

so LGTM

(*) only change required: in Makefile.builder
```
-       opam depext -i -y mirage.3.8.0 lwt.5.3.0
+       opam depext -i -y mirage
```
(which is basicly the same as one of your last changes to the Dockerfile)

want a separate PR?

PR #117, opened 2020-10-24
README: use kernelopts='' instead of None

hannesm:
this is especially important with the upcoming mirage 3.9 release (which no longer filters boot parameters).

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Run this command in dom0 to create a `mirage-firewall` VM using the `mirage-fire
 ```
 qvm-create \
   --property kernel=mirage-firewall \
-  --property kernelopts=None \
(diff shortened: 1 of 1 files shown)


talex5:
Thanks - I guess that explains this too: https://github.com/mirage/qubes-mirage-firewall/pull/116#issuecomment-677485127

hannesm:
@talex5 yes, I was looking for the command line on qubes to set the boot parameters to add to the changes for mirage 3.9.0, when I re-read the comments in #116 and looked into README.

hannesm:
according to @marmarek in https://github.com/mirage/mirage/issues/1159#issuecomment-656201463 we may need `qvm-features VMNAME no-default-kernelopts 1` as well?

Issue #120, opened 2020-10-29
With solo5, firewall runs out of memory after a while and can't recover


talex5:
First reported by @palainp in https://github.com/mirage/qubes-mirage-firewall/pull/116#issuecomment-708936007. I see this too now, after a few days.

```
2020-10-26 15:26:48 -00:00: INF [memory_pressure] Writing meminfo: free 115MiB / 124MiB (93.24 %)
[...]
2020-10-27 09:00:57 -00:00: INF [client_eth:dom33:10.137.0.15] who-has 10.137.0.16? responding with fe:ff:ff:ff:ff:ff
2020-10-27 09:01:04 -00:00: INF [memory_pressure] Writing meminfo: free 13MiB / 124MiB (9.99 %)
2020-10-27 09:01:04 -00:00: WRN [firewall] Memory low - dropping packet and resetting NAT table
2020-10-27 09:01:04 -00:00: INF [memory_pressure] Writing meminfo: free 13MiB / 124MiB (9.99 %)
2020-10-27 09:01:04 -00:00: WRN [firewall] Memory low - dropping packet and resetting NAT table
2020-10-27 09:01:04 -00:00: INF [memory_pressure] Writing meminfo: free 13MiB / 124MiB (9.99 %)
2020-10-27 09:01:04 -00:00: WRN [firewall] Memory low - dropping packet and resetting NAT table
```

hannesm:
I managed to get memtrace (and memtrace_viewer) to work with MirageOS unikernels -- take a look at https://github.com/hannesm/memtrace-mirage this may be useful for debugging it further (being able to see the OCaml view of which call stacks refer to how much memory) -- see https://blog.janestreet.com/finding-memory-leaks-with-memtrace/ for a great description of memtrace_viewer.

palainp:
My current thinking is that ```malloc_footprint()``` is too pessimistic and need some time to be updated when we run out of free memory.

The second solution offered by @mato in https://github.com/mirage/qubes-mirage-firewall/pull/116#issuecomment-704827905 was to use ```sbrk(0)``` instead of ```malloc_footprint()``` but this function is currently not available in nolibc (https://github.com/mirage/ocaml-freestanding/blob/master/nolibc/include/unistd.h).

[...]

PR #121, opened 2020-10-29
Report memory usage every 10 minutes

talex5:
Might help with debugging https://github.com/mirage/qubes-mirage-firewall/issues/120


talex5:
As @winux138 points out, this code is completely wrong because it just keeps printing the same stats!

PR #122, opened 2020-11-12
Update README.md

dalrrard:
Added a small bit of troubleshooting for the docker service if it fails to come up. This occurred while using the home directory symlink in a Fedora 32 VM.


dalrrard:
This looks to be a more general problem that just using Fedora. The option you need to change is different depending on which version you run and if you're running in Fedora or Debian though. Probably not good to have on the main install page.

Issue #134, opened 2021-10-22
Qubes 4.1rc1 error - "Fatal trap"


ArtificialAmateur:
Building from source from the current master HEAD gives the following error when deployed and ran:
```
[2021-10-22 14:14:41] Solo5: Xen console: port 0x2, ring @0x00000000FEFFF000
[2021-10-22 14:14:41]             |      ___|
[2021-10-22 14:14:41]   __|  _ \  |  _ \ __ \
[2021-10-22 14:14:41] \__ \ (   | | (   |  ) |
[2021-10-22 14:14:41] ____/\___/ _|\___/____/
[2021-10-22 14:14:41] Solo5: Bindings version v0.6.7
[2021-10-22 14:14:41] Solo5: Memory map: 64 MB addressable:
[2021-10-22 14:14:41] Solo5:   reserved @ (0x0 - 0xfffff)
[2021-10-22 14:14:41] Solo5:       text @ (0x100000 - 0x2effff)
[2021-10-22 14:14:41] Solo5:     rodata @ (0x2f0000 - 0x353fff)
[2021-10-22 14:14:41] Solo5:       data @ (0x354000 - 0x4e7fff)
[2021-10-22 14:14:41] Solo5:       heap >= 0x4e8000 < stack < 0x4000000
[2021-10-22 14:14:41] 2021-10-22 18:14:41 -00:00: INF [qubes.rexec] waiting for client...
[2021-10-22 14:14:41] 2021-10-22 18:14:41 -00:00: INF [qubes.gui] waiting for client...
[2021-10-22 14:14:41] 2021-10-22 18:14:41 -00:00: INF [qubes.db] connecting to server...
[2021-10-22 14:14:41] 2021-10-22 18:14:41 -00:00: INF [qubes.db] connected
[2021-10-22 14:14:41] 2021-10-22 18:14:41 -00:00: INF [qubes.rexec] client connected, other end wants to use protocol version 3, continuing with version 2
[2021-10-22 14:14:41] Solo5: trap: type=#MF ec=0x0 rip=0x28bded rsp=0x3ffd600 rflags=0x10297
[2021-10-22 14:14:41] Solo5: ABORT: cpu_x86_64.c:181: Fatal trap
```

Reopening issue #128.

palainp:
Can you try to run in your compilation VM `objdump -D qubes_firewall.xen` and then go down to` 28bded`?
[...]

PR #141, opened 2022-08-13
Drop PV references from README.md

palainp:
I think it might be useful to remove the PV references in the README.md because since mirage 3.9 it only compiles for PVH VMs (#116), and versions prior to v0.8.0 of ocaml-solo5 cannot run under Qubes 4.1 (#134).

As Qubes 4.0 is now EOL, any user must use qubes-mirage-firewall with mirage 4.0 and a PVH VM.


hannesm:
thanks. this looks good :)

PR #142, opened 2022-08-30
firewall rule: remove DNS rule (was only needed in Qubes 3)

### LINKED ISSUE #63: Remove DNS rule?
In Qubes 3, AppVMs were configured to send DNS traffic to the firewall, which redirected it to sys-net. In Qubes 4 it looks like AppVMs are configured to use `10.139.1.{1,2}` as their nameservers and an iptables rule in the NAT table of sys-net redirects it. Do we still need the DNS redirection rule in mirage-firewall?

Strangely, on my system I can do DNS lookups on any IP at all and it works. e.g. `nslookup www.google.com 200.123.123.123` works, even in sys-net. I don't know what is doing that.

hannesm:
//cc @palainp would you mind to try this out?

fixes #63

### DIFF
--- a/client_net.ml
+++ b/client_net.ml
@@ -59,7 +59,7 @@ let input_arp ~fixed_arp ~iface request =
       iface#writev `ARP (fun b -> Arp_packet.encode_into response b; Arp_packet.size)
 
 (** Handle an IPv4 packet from the client. *)
-let input_ipv4 get_ts cache ~iface ~router dns_client packet =
+let input_ipv4 get_ts cache ~iface ~router dns_client dns_servers packet =
   let cache', r = Nat_packet.of_ipv4_packet !cache ~now:(get_ts ()) packet in
   cache := cache';
   match r with
@@ -70,7 +70,7 @@ let input_ipv4 get_ts cache ~iface ~router dns_client packet =
   | Ok (Some packet) ->
     let `IPv4 (ip, _) = packet in
     let src = ip.Ipv4_packet.src in
-    if src = iface#other_ip then Firewall.ipv4_from_client dns_client router ~src:iface packet
+    if src = iface#other_ip then Firewall.ipv4_from_client dns_client dns_servers router ~src:iface packet
     else (
       Log.warn (fun f -> f "Incorrect source IP %a in IP packet from %a (dropping)"
                    Ipaddr.V4.pp src Ipaddr.V4.pp iface#other_ip);
@@ -78,7 +78,7 @@ let input_ipv4 get_ts cache ~iface ~router dns_client packet =
     )
 
 (** Connect to a new client's interface and listen for incoming frames and firewall rule changes. *)
-let add_vif get_ts { Dao.ClientVif.domid; device_id } dns_client ~client_ip ~router ~cleanup_tasks qubesDB =
+let add_vif get_ts { Dao.ClientVif.domid; device_id } dns_client dns_servers ~client_ip ~router ~cleanup_tasks qubesDB =
   Netback.make ~domid ~device_id >>= fun backend ->
   Log.info (fun f -> f "Client %d (IP: %s) ready" domid (Ipaddr.V4.to_string client_ip));
   ClientEth.connect backend >>= fun eth ->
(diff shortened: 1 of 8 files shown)


hannesm:
hmm, I'm even more curious that rules.ml still has some default_dns_servers hardcoded -- is this and the special dns rule still a thing with Qubes 4.x?

palainp:
This is compiling and running fine (with docker the sha256 is `1855cb82fce272d93c43ecf06d175ae82e07c309855fe8cd5fdd6313a7f54a0f`).

I managed to run a speedtest and this was ok (the bandwidth is still less than a linux sys-firewall, around 60%, the cpu seems to be a possible bottleneck, I have to investigate that later).

The IPs (`10.139.1.1` and `10.139.1.2`) are those provided by Qubes and seems to be translated to the next AppVM in the network chain:
```
$ cat /etc/resolv.conf 
nameserver 10.139.1.1
nameserver 10.139.1.2
```
I'm not sure if we can get them automatically in `uplink.ml`?

palainp:
Thinking back, and as the fw shouldn't have to resolve something, I tried to remove that part of the code:
```
diff --git a/rules.ml b/rules.ml
index f72d6c0..541d621 100644
--- a/rules.ml
+++ b/rules.ml
@@ -10,12 +10,6 @@ module Q = Pf_qubes.Parse_qubes
 let src = Logs.Src.create "rules" ~doc:"Firewall rules"
 module Log = (val Logs.src_log src : Logs.LOG)
 
-(* the upstream NetVM will redirect TCP and UDP port 53 traffic with
-   these destination IPs to its upstream nameserver. *)
-let default_dns_servers = [
-  Ipaddr.V4.of_string_exn "10.139.1.1";
-  Ipaddr.V4.of_string_exn "10.139.1.2";
-]
 let dns_port = 53
 
 module Classifier = struct
@@ -26,14 +20,6 @@ module Classifier = struct
 
   let matches_proto rule packet = match rule.Q.proto, rule.Q.specialtarget with
     | None, None -> true
-    | None, Some `dns when List.mem packet.ipv4_header.Ipv4_packet.dst default_dns_servers -> begin
-      (* specialtarget=dns applies only to the specialtarget destination IPs, and
-         specialtarget=dns is also implicitly tcp/udp port 53 *)
-      match packet.transport_header with
-        | `TCP header -> header.Tcp.Tcp_packet.dst_port = dns_port
-        | `UDP header -> header.Udp_packet.dst_port = dns_port
-        | _ -> false
-      end
    (* DNS rules can only match traffic headed to the specialtarget hosts, so any other destination
    isn't a match for DNS rules *)
     | None, Some `dns -> false
```
and it still works (but probably needs a bit more testing to make sure you don't miss any corner cases).

hannesm:
Thanks @palainp. So from my understanding, the specialtarget `` `dns `` should match for dns packets:
(a) in doa.ml (read_network_config) we extract the `/qubes-primary-dns` already (but there's eventually a `/qubes-secondary-dns`?)
(b) we should use these value instead of `Rules.default_dns_servers` to check..

Certainly the entire `` `dns `` can be removed -- especially if you don't have any special firewall rules for `` `dns `` in your setup. But the semantics in the qubes firewall ruleset is likely that it should respect the QubesDB values.

hannesm:
I added a commit here which removes the hardcoded IPv4 addresses, and uses those from QubesDB (read once at startup). Could you give this a try, @palainp?

palainp:
It compiles and starts like before, but I somehow got:
```
[2022-09-07 17:11:30] Raised at Netchannel__Frontend.Make.write.(fun) in file "duniverse/mirage-net-xen/lib/frontend.ml", line 377, characters 16-24
[2022-09-07 17:11:30] Called from Lwt.Resolution_loop.handle_with_async_exception_hook in file "duniverse/lwt/src/core/lwt.ml", line 1144, characters 8-11
[2022-09-07 17:11:30] Solo5: solo5_exit(2) called
```
Not sure if that is related, but I never got that. I try to reproduce.

hannesm:
hmm, strange...

palainp:
It's certainly not related (see another appearance of `Netif failure` https://github.com/mirage/qubes-mirage-firewall/issues/120#issuecomment-932711576 ). I can't reproduce it and it works as before now.
It's ok for me to be merged!

hannesm:
merged as part of #149

Issue #143, opened 2022-09-05
Random qvm shutdown


kennethrrosen:
I have set my mirage-firewall VM as the NetVM for each of my networked VMs. After sometime, the VM randomly shutsdown without warning. This happens rather consistently. I have the VM to start on boot. Once I restart the mirage-firewall VM, it generally doesn't crash again.

palainp:
Hi, thanks for your feedback, I've had something like this only once over the last few weeks and I couldn't reproduce it.

I currently have the same configuration: mirage-fw is the default netvm and is run at boot time. I had the crash when I lost my WiFi connection and I guess mirage-fw couldn't free up enough memory with packets still coming in, so the heap grows until mirage-fw fails to get fresh memory from solo5.

In order to validate this can you check in your log if you have a message like (`grep -C1 "Aborted" /var/log/xen/console/guest-mirage-test.log`):
```
Fatal error: out of memory
Aborted
Solo5: solo5_abort() called
```
The interesting part should be just above, my logs only shows the ARP message about a minute before.

I can see two easy tweaks to try to fix that:
- increase the memory given to mirage-fw (I don't like this because the low memory requirement is one of the goals of using mirage-fw)
- change the percentage of free memory when forcing the GC call (in https://github.com/mirage/qubes-mirage-firewall/blob/b0205f7dab9d7af5a0a2cdbd90fef10aeaf6cc07/memory_pressure.ml#L44 . After the GC call if we still have a less than 60% of free memory we may raise a `Memory_critical` to drop the NAT table in https://github.com/mirage/qubes-mirage-firewall/blob/b0205f7dab9d7af5a0a2cdbd90fef10aeaf6cc07/firewall.ml#L86)

If you're ok with the build system, can you try the test-memory branch on my repository?

kennethrrosen:
Apologies for the late reply. With a more detailed walkthrough I could try the test-memory branch, but for now I've slightly increased the memory given to mirage-fw. There have been no further crashes.

hannesm:
I'll close this issue, thanks for your information. If you have an answer or @palainp questions, don't hesitate to comment here :)

burghardt:
I catch this today. I experienced random aborts of firewall VM for few days. Looks like this is related to moderately heavy traffic (updating templates, opening multiple websites in tabs, etc.).
```
Fatal error: out of memory
Aborted
Solo5: solo5_abort() called
```
[...]

PR #147, opened 2022-09-13
Test memory

hannesm:
see dscussion in #143 /cc @palainp


palainp:
Thanks @hannesm for this PR.

With the change to `Xen_os.quick_stat()`, I expected to have some adjustments to make here. Do you think it might be useful to try `GC.compact()` instead of just `GC.full_major()` when the memory pressure test is triggered?

hannesm:
Hmm, do you mean `Gc.compact` is faster? It may be sufficient -- the main thing is that the major cycle finishes and finalizers are executed (to free up bigarray data) -- as far as I know.

At the same time, unless I'm wrong we're using the (default since 4.13) best-fit GC strategy, where compactions are not needed (but in any way, both Gc.compact and Gc.full_major do the major collection -- I'm not entirely sure what the difference is) -- do you observe speed-wise a huge difference?

palainp:
I haven't noticed any speed difference and the documentation warns that `Gc.compact` is slower (https://v2.ocaml.org/api/Gc.html#VALcompact). Thanks for reminding me that best-fit is now the default and that there is no need to compact, it's probably not worth investigating so much that idea.

hannesm:
merged as part of #149

PR #144, opened 2022-09-07
remove gui code, not needed anymore in Qubes 4.1

palainp:
This PR should fix #62 :)

PR #149, opened 2022-09-14
Next release

hannesm:
//cc @palainp this is a combination of #147 #142 #144, and also updates the opam-repository to mirage-qubes 0.9.3. I updated the sha256.

Would you mind to see whether you get the same result, and if yes, I'd be keen on merging this and cut a fresh 0.8.1 release.


palainp:
It's good to me! Thanks for the next release.
I got the same sha256, and the firewall starts correctly+support load as before.

PR #150, opened 2022-10-06
remove frameQ

palainp:
This update mainly removes code that is no longer needed:
- frameQ was used to limit the number of promises waiting to be delivered, and somehow it now works without that + it increases performance with tcp as it doesn't drop packets (which will have to be retransmitted)
- the nat table should no longer be dropped (and is capped at a lower value by default) as it should not grow without control these days (but we still need to call memory_pressure.status to GC from time to time)
- the firewall seems to have decent performance (according to local iperf tests) with only 32MB


config.ml
```diff
@@ -10,7 +10,7 @@ let table_size =
       ~doc:"The number of NAT entries to allocate."
       ~docv:"ENTRIES" ["nat-table-size"]
   in
-  let key = Key.Arg.opt ~stage:`Both Key.Arg.int 5_000 info in
+  let key = Key.Arg.opt ~stage:`Both Key.Arg.int 500 info in
```
hannesm, line 13: I suggest to remove this change from the patch.

hannesm:
In my_nat.ml, should we as well apply the diff:
```OCaml
diff --git a/my_nat.ml b/my_nat.ml
index 2652ff5..94cf790 100644
--- a/my_nat.ml
+++ b/my_nat.ml
@@ -64,15 +64,7 @@ let remove_connections t ports ip =
   ports.nat_icmp := Ports.diff !(ports.nat_icmp) (Ports.of_list freed_ports.Mirage_nat.icmp)
 
 let add_nat_rule_and_translate t ports ~xl_host action packet =
-  let apply_action xl_port =
-    Lwt.catch (fun () ->
-        Nat.add t.table packet (xl_host, xl_port) action
-      )
-      (function
-        | Out_of_memory -> Lwt.return (Error `Out_of_memory)
-        | x -> Lwt.fail x
-      )
-  in
+  let apply_action xl_port = Nat.add t.table packet (xl_host, xl_port) action in
   let rec aux ~retries =
     let nat_ports, dns_ports =
       match packet with
@@ -82,12 +74,6 @@ let add_nat_rule_and_translate t ports ~xl_host action packet =
     in
     let xl_port = pick_free_port ~nat_ports ~dns_ports in
     apply_action xl_port >>= function
-    | Error `Out_of_memory ->
-      (* Because hash tables resize in big steps, this can happen even if we have a fair
-         chunk of free memory. *)
-      Log.warn (fun f -> f "Out_of_memory adding NAT rule. Dropping NAT table...");
-      reset t ports >>= fun () ->
-      aux ~retries:(retries - 1)
     | Error `Overlap when retries < 0 -> Lwt.return (Error "Too many retries")
     | Error `Overlap ->
       if retries = 0 then (
```

config.ml
palainp: Ok I will! Thanks!

hannesm:
the proposed my_nat changes above are done in #151

hannesm:
thanks, merged via #152

PR #154, opened 2022-10-27
update to dns 6.4.0

hannesm:
There's an API change in dns-client (the function `connect` returns a tuple of protocols and flow), which we follow with this commit.


palainp:
It compiles and run fine here. I also have the same hash256. Ok to merge for me :)

Issue #158, opened 2022-11-08
DNS pb with v0.8.2


palainp:
First reported by fiftyfourthparallel on https://forum.qubes-os.org/t/mirage-firewall-0-8-2-broken-new-users-should-install-0-8-1/14566
It seems that we now have an issue with DNS. I'll try to reproduce the scenario to understand what is the pb root cause.

palainp:
I have the following in my test VM:
```
$ qvm-firewall test list
NO  ACTION  HOST            PROTOCOL  PORT(S)  SPECIAL TARGET  ICMP TYPE  EXPIRE  COMMENT
0   accept  www.google.com  tcp       443      -               -          -       -
1   accept  -               -         -        dns             -          -       -
2   accept  -               icmp      -        -               -          -       -
3   drop    -               -         -        -               -          -       -
```
I can reproduce the crash that with the 0.8.2 binary and with the current robur.coop build. I'll try to figure out what is the issue.

palainp:
It seems to be stuck resolving: https://github.com/mirage/qubes-mirage-firewall/blob/b09acdeec2d91b8b9767f39b5752de3addd8f17d/rules.ml#L54

hannesm:
Would you mind to compile and try the `debug-dns` branch of my fork (hannesm/qubes-mirage-firewall.git)? and provide the output here...

palainp:
You spotted the right thing, it loops with:
```
[2022-11-08 18:15:03] 2022-11-08 17:15:03 -00:00: INF [my-nat] nat says UDP port 41277 is not free
[2022-11-08 18:15:03] 2022-11-08 17:15:03 -00:00: INF [my-nat] trying UDP port 1739
[2022-11-08 18:15:03] 2022-11-08 17:15:03 -00:00: INF [my-nat] nat says UDP port 1739 is not free
[2022-11-08 18:15:03] 2022-11-08 17:15:03 -00:00: INF [my-nat] trying UDP port 58504
[2022-11-08 18:15:03] 2022-11-08 17:15:03 -00:00: INF [my-nat] nat says UDP port 58504 is not free
[2022-11-08 18:15:03] 2022-11-08 17:15:03 -00:00: INF [my-nat] trying UDP port 48392
[2022-11-08 18:15:03] 2022-11-08 17:15:03 -00:00: INF [my-nat] nat says UDP port 48392 is not free
[...]

PR #159, opened 2022-11-08
My_nat.free_udp_port: avoid looping forever, use last_resort_port earlier

hannesm:
This avoids using the entire CPU. See #158 for the issue description (especially my comment https://github.com/mirage/qubes-mirage-firewall/issues/158#issuecomment-1307610459).

### DIFF
--- a/my_nat.ml
+++ b/my_nat.ml
@@ -46,22 +46,25 @@ let pick_free_port t proto =
   go 10
 
 let free_udp_port t ~src ~dst ~dst_port =
-  let rec go () =
-    let src_port =
-      Option.value ~default:t.last_resort_port (pick_free_port t `Udp)
-    in
-    if Nat.is_port_free t.table `Udp ~src ~dst ~src_port ~dst_port then begin
-      let remove =
-        if src_port <> t.last_resort_port then begin
-          t.udp_dns <- S.add src_port t.udp_dns;
-          (fun () -> t.udp_dns <- S.remove src_port t.udp_dns)
-        end else Fun.id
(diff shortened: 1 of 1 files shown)


palainp:
Thanks for that PR. After some times setting up correctly my mirage-fw I can confirm that it doesn't burn the CPU anymore.
LGTM (together with the mirage-nat update that also helps in solving this issue)!

hannesm:
I would like to verify first that the mirage-nat change (and 3.0.1 release) fixes the issue. This change here is "purely cosmetical" and shouldn't be needed. Though it as well helps to avoid busy loops (since the `free_udp_port` will be bounded).

While working on this, another issue is how DNS replies are handled with a single global mailbox (Lwt_mvar)
- a DNS query is sent out
- whenever a DNS reply is received, this is pushed into the mailbox

This means there can be concurrent requests with unclear ordering, and also the Lwt_mvar.put (in uplink) can block if there's nobody reading the mailbox (which only my_dns is). So that code should be revised that DNS replies are matching the requests, and there's no blocking behaviour if more replies are received than requests that were sent. Especially evil since there's a timeout which cancels the Lwt_mvar.take...

palainp:
Thanks @hannesm for your amazing work on that points!
I can confirm that the PR with mirage-nat alone (without this PR) permit to not overload CPU, but I'll feel better to have a bounded loop :)

hannesm:
this has been part of https://github.com/mirage/qubes-mirage-firewall/pull/163 and the 0.8.3 release

Issue #155, opened 2022-10-27
mirage vm crash on downstream vm start


xaki23:
this is most likely the same issue as https://github.com/mirage/qubes-mirage-firewall/issues/35 but i am creating a fresh one instead of reopening because the backtrace is very different due to all the changes that happened in the last 4 (!) years.

repro steps:
* create a mirage fw vm
* connect some downstream vm
* restart the downstream vm 64 times
* mirage fw crashes

```
[2022-10-27 21:14:44] 2022-10-27 19:14:43 -00:00: INF [client_net] add client vif {domid=3091;device_id=0} with IP 10.137.0.90
[2022-10-27 21:14:44] Fatal error: exception Xs_protocol.Error("E2BIG")
[2022-10-27 21:14:44] Raised at Xs_protocol.response in file "duniverse/ocaml-xenstore/core/xs_protocol.ml", line 685, characters 13-28
[2022-10-27 21:14:44] Called from Xs_client_lwt.Client.rpc.(fun) in file "duniverse/ocaml-xenstore/client_lwt/xs_client_lwt.ml", line 318, characters 13-50
[2022-10-27 21:14:44] Called from Lwt.Sequential_composition.bind.create_result_promise_and_callback_if_deferred.callback in file "duniverse/lwt/src/core/lwt.ml", line 1849, characters 23-26
[2022-10-27 21:14:44] Re-raised at Lwt.Miscellaneous.poll in file "duniverse/lwt/src/core/lwt.ml", line 3077, characters 20-29
[2022-10-27 21:14:44] Called from Xen_os__Main.run.aux in file "duniverse/mirage-xen/lib/main.ml", line 37, characters 10-20
[2022-10-27 21:14:44] Called from Dune__exe__Main.run in file "main.ml" (inlined), line 3, characters 12-29
[2022-10-27 21:14:44] Called from Dune__exe__Main in file "main.ml", line 115, characters 5-10
[2022-10-27 21:14:44] Solo5: solo5_exit(2) called
```

the E2BIG right after add-client-vif suggests its still the "xenstore subscriptions leaking" issue.
see https://github.com/mirage/qubes-mirage-firewall/issues/35 and the various issues linked from it.

palainp:
Hi, thanks for this valuable reproducible protocol!
I can see that too now, and I added some logs in `ocaml-xenstore/.../xs_client_lwt.ml`. The following shows what s watched during a `qvm-start vm && qvm-kill vm`:
```
...
[...]

PR #160, opened 2022-11-10
use a fresh client for requesting vif and ip

hannesm:
in the callback to "Xs_client.wait", all operations are tracked and new watches are installed (that are never removed, due to xenstore's xs_handle "accessed_path" never removes any elements of the "accessed_paths" (a mutable StringSet). So, whatever is done in the callback of wait needs to take care (if returning EAGAIN and thus forcing xenstore to continue waiting/watching) that accesses are tracked.

Our way out is to create a fresh client and read the IP address with that new client -> the watcher isn't extended -> no dangling (leaking) watches, and no leaking only-expanding StringSet.

tentative fix for #155

### DIFF
--- a/dao.ml
+++ b/dao.ml
@@ -65,43 +65,44 @@ let read_rules rules client_ip =
                              icmp_type = None;
                              number = 0;})]
 
-let vifs ~handle domid =
+let vifs client domid =
   match String.to_int domid with
   | None -> Log.err (fun f -> f "Invalid domid %S" domid); Lwt.return []
   | Some domid ->
     let path = Printf.sprintf "backend/vif/%d" domid in
-    directory ~handle path >>=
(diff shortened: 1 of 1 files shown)


hannesm:
diff best viewed without whitespace changes, https://github.com/mirage/qubes-mirage-firewall/pull/160/files?w=1

palainp:
Thanks again @hannesm for your time and your hard work on this! I also tried to think how to handle the watches in xenstore but also failed.

With this PR I was able to successively start & stop 100+ clients, I think it solve the issue so LGTM so far.

@xaki23 would you mind to test that on your setup too?

hannesm:
this has been part of https://github.com/mirage/qubes-mirage-firewall/pull/163 and the 0.8.3 release

PR #162, opened 2022-11-11
DNS: start task reading Lwt_mvar and distributing DNS replies to clients

hannesm:
Before, a DNS request was sent and the first thing appearing in the Lwt_mvar was taken as reply. The issue with this was two-fold:
- it could be a reply for a different request
- there could be DNS replies being sent to the uplink stack leading to Lwt_mvar.put being called, which blocks if there is already a value in the mvar.

No, the separate task is a loop reading the mvar, using a Lwt_condition to signal the receive of that ID (potentially discarding if there's no client waiting). The DNS query registers itself (using the ID) in the map with a Lwt_condition, and waits to be notified (or a timeout occurs).

as outlined in #159


palainp:
It works for me, I think we can merge it! Thanks for that @hannesm!

hannesm:
this has been part of #163 and the 0.8.3 release

PR #164, opened 2022-11-13
Reproducible build systems: use in GitHub action the build-with-docker.sh

hannesm:
Also upload the artifact to GitHub action, and in addition use the same setup (ubuntu 20.04 image) and build directories as done on builds.robur.coop.

Also use `strip` on the resulting binary to reduce it's size (since the debug section aren't mapped into the running unikernel, there's nothing we get from them -- also they are preserved (as .debug file) and uploaded to https://builds.robur.coop if one needs them).

This entails binary reproducibility between the different systems:
- a developer using ./build-with-docker.sh
- GitHub action (run on every PR)
- builds.robur.coop with the ubuntu-20.04 worker

The situation before:
- build-with-docker.sh used a Docker image from "ocaml/opam" which was based on fedora-35 and included OCaml 4.14.
- GitHub action used the ubuntu-latest (a 22.04 image, at the current state of writing https://github.com/actions/runner-images)
- builds.robur.coop used a ubuntu 20.04 image https://builds.robur.coop/job/qubes-firewall?platform=ubuntu-20.04

I tried to get them reproducible, but even using a "ubuntu 22.04" image from ocaml/opam and GitHub action did not lead to the same output (it is the same C compiler, but they report versions differently for some reason).

So instead I stick to the official ubuntu 20.04 image (we can of course upgrade at any time -- but we should then as well have a worker on builds.robur.coop doing this build), which is very small in size.

The drawback is that compilation (execution of ./build-with-docker.sh) on a high-speed Internet connection will take some more time (since the OCaml compiler is bootstrapped). Another drawback is that in the Dockerfile the `opam` binary is manually downloaded from github.com -- reason is that we need a recent one (>= 2.1.0 for MirageOS 4; and ubuntu 20.04 only ships 2.0.5 -- solved in ubuntu 22.04 which ships 2.1.2).

The advantage is that we've build information (reproducible-builds.org terminology) / SBOM (software bill of materials) data independent of Docker:
- the system packages https://builds.robur.coop/job/qubes-firewall/build/04fa553a-a4dd-4433-b853-f91e6fa0d12e/f/system-packages
- the opam packages (with checksums or git commits) https://builds.robur.coop/job/qubes-firewall/build/04fa553a-a4dd-4433-b853-f91e6fa0d12e/f/opam-switch

And any user of qubes-mirage-firewall can lookup their binary sha256 (sha256 /path/to/vmlinuz ; with this PR merged it is 3f71a1b672a15d145c7d40405dd75f06a2b148d2cfa106dc136e3da38552de41) at https://builds.robur.coop/hash?sha256=3f71a1b672a15d145c7d40405dd75f06a2b148d2cfa106dc136e3da38552de41 (there's an input field on https://builds.robur.coop) and find out
- the build information (sources [system packages, opam packages]) sufficient to reproduce the exact same binary (eventually even on other Linux distributions when using the same C compiler, etc.)
- whether there are newer builds with different output (and what changed in terms of input, e.g. a new release of a dependency)

All this is in my opinion a much neater story for the binaries we provide. NB that builds.robur.coop uses (a) orb (from https://github.com/roburio/orb) for building the package and (b) on a daily basis the HEAD of opam-repository (while `Dockerfile` has a specific commit hardcoded). This means that the latest build on builds.robur.coop may be different from the local build or GitHub action build.

//cc @palainp

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,8 +14,6 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
-        ocaml-compiler:
-          - 4.14.x
 
     runs-on: ${{ matrix.os }}
 
@@ -23,19 +21,10 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
-      - name: Use OCaml ${{ matrix.ocaml-compiler }}
-        uses: ocaml/setup-ocaml@v2
+      - run: ./build-with-docker.sh
+
+      - name: Upload Artifact
+        uses: actions/upload-artifact@v3
         with:
-          ocaml-compiler: ${{ matrix.ocaml-compiler }}
-
-      - run: opam depext solo5 "mirage>4"
-
-      - run: opam install solo5 "mirage>4"
-
-      - run: opam exec -- mirage configure -t xen
-
-      - run: opam exec -- make depend
-
-      - run: opam exec -- dune build
-
-      - run: sha256sum dist/qubes-firewall.xen
+          name: mirage-firewall.tar.bz2
+          path: mirage-firewall.tar.bz2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,6 +23,8 @@ jobs:
 
       - run: ./build-with-docker.sh
 
+      - run: sh -exc 'if [ $(sha256sum dist/qubes-firewall.xen | cut -d " " -f 1) = $(grep "SHA2 last known" build-with-docker.sh | rev | cut -d ":" -f 1 | rev | cut -d "\"" -f 1 | tr -d " ") ]; then echo "SHA256 MATCHES"; else exit 42; fi'
+
       - name: Upload Artifact
         uses: actions/upload-artifact@v3
         with:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,21 @@
 # Pin the base image to a specific hash for maximum reproducibility.
 # It will probably still work on newer images, though, unless an update
 # changes some compiler optimisations (unlikely).
-# fedora-35-ocaml-4.14
-FROM ocaml/opam@sha256:68b7ce1fd4c992d6f3bfc9b4b0a88ee572ced52427f0547b6e4eb6194415f585
-ENV PATH="${PATH}:/home/opam/.opam/4.14/bin"
+# ubuntu-20.04
+FROM ubuntu@sha256:b25ef49a40b7797937d0d23eca3b0a41701af6757afca23d504d50826f0b37ce
 
-# Since mirage 4.2 we must use opam version 2.1 or later
-RUN sudo ln -sf /usr/bin/opam-2.1 /usr/bin/opam
+RUN apt update && apt install --no-install-recommends --no-install-suggests -y wget ca-certificates git patch unzip make gcc g++ libc-dev
+RUN wget -O /usr/bin/opam https://github.com/ocaml/opam/releases/download/2.1.3/opam-2.1.3-i686-linux && chmod 755 /usr/bin/opam
 
+ENV OPAMROOT=/tmp
+ENV OPAMCONFIRMLEVEL=unsafe-yes
 # Pin last known-good version for reproducible builds.
 # Remove this line (and the base image pin above) if you want to test with the
 # latest versions.
-RUN cd /home/opam/opam-repository && git fetch origin master && git reset --hard 685eb4efcebfa671660e55d76dea017f00fed4d9 && opam update
-
-RUN opam install -y mirage opam-monorepo ocaml-solo5
-RUN mkdir /home/opam/qubes-mirage-firewall
-ADD config.ml /home/opam/qubes-mirage-firewall/config.ml
-WORKDIR /home/opam/qubes-mirage-firewall
-CMD opam exec -- mirage configure -t xen && make depend && make tar
+RUN opam init --disable-sandboxing -a --bare https://github.com/ocaml/opam-repository.git#685eb4efcebfa671660e55d76dea017f00fed4d9
+RUN opam switch create myswitch 4.14.0
+RUN opam exec -- opam install -y mirage opam-monorepo ocaml-solo5
+RUN mkdir /tmp/orb-build
+ADD config.ml /tmp/orb-build/config.ml
+WORKDIR /tmp/orb-build
+CMD opam exec -- sh -exc 'mirage configure -t xen --allocation-policy=best-fit && make depend && make tar'
--- a/Makefile.user
+++ b/Makefile.user
@@ -1,6 +1,8 @@
 tar: build
 	rm -rf _build/mirage-firewall
 	mkdir _build/mirage-firewall
+	cp dist/qubes-firewall.xen dist/qubes-firewall.xen.debug
+	strip dist/qubes-firewall.xen
 	cp dist/qubes-firewall.xen _build/mirage-firewall/vmlinuz
 	touch _build/mirage-firewall/modules.img
 	cat /dev/null | gzip -n > _build/mirage-firewall/initramfs
--- a/build-with-docker.sh
+++ b/build-with-docker.sh
@@ -3,7 +3,7 @@ set -eu
 echo Building Docker image with dependencies..
 docker build -t qubes-mirage-firewall .
 echo Building Firewall...
-docker run --rm -i -v `pwd`:/home/opam/qubes-mirage-firewall qubes-mirage-firewall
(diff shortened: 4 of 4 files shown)


palainp:
This is awesome! Thanks!

The bootstrap for install ocaml 4.14 is a bit annoying but I think the benefit for having reproducibility worth it :)
May be it is possible to increase the cache usage and avoid sha256 mismatch by removing the `apt update`? (to me if a package is updated in the install list is updated repositories it may change the sum, esp. gcc for the solo5 part).

palainp:
And btw I have the same shasum, and it runs well, so LGTM!

hannesm:
@palainp suggested
> May be it is possible to avoid sha256 mismatch by removing the apt update?

That sounds like a good suggestion, unfortunately I get the following errors:
```shell
Step 2/12 : RUN apt install --no-install-recommends --no-install-suggests -y wget ca-certificates git patch unzip make gcc g++ libc-dev
 ---> Running in 5988b7c42f30

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

Reading package lists...
Building dependency tree...
Reading state information...
Package ca-certificates is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Unable to locate package wget
E: Package 'ca-certificates' has no installation candidate
E: Unable to locate package git
E: Unable to locate package patch
E: Unable to locate package unzip
E: Unable to locate package make
E: Unable to locate package gcc
E: Unable to locate package g+
E: Unable to locate package libc-dev
The command '/bin/sh -c apt install --no-install-recommends --no-install-suggests -y wget ca-certificates git patch unzip make gcc g++ libc-dev' returned a non-zero code: 100
```

Since I have barely any knowledge about "docker" and what guarantees Linux distribution give (esp. ubuntu with LTS) in terms of packaging and updates, I don't know how to proceed. I can see how the Dockerfile as suggested here will fail (with newer gcc releases, etc.).

Taking the other path, maybe we need to pick / pin the set of system packages that should be installed explicitly (i.e. name and version) -- as done by orb / builds.robur.coop in the system-packages file (which can be fed to `cat filename | xargs apt-get install -y`)?

As alternative we can push intermediate docker containers (somewhere?) and have them be used by the script, but then certainly two more issues are present: how to reproduce that container and who is in charge of updating it?

palainp:
Oh that's bad. I should have tried out myself before proposing that, sorry :/
To help I think we can update the Dockerfile like that:
```
diff --git a/Dockerfile b/Dockerfile
index c511cdb..56ca5f3 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,8 @@
 # ubuntu-20.04
 FROM ubuntu@sha256:b25ef49a40b7797937d0d23eca3b0a41701af6757afca23d504d50826f0b37ce
 
-RUN apt update && apt install --no-install-recommends --no-install-suggests -y wget ca-certificates git patch unzip make gcc g++ libc-dev
+RUN apt-get update && apt-get install --no-install-recommends --no-install-suggests -y wget ca-certificates git patch unzip make \
+gcc='4:9.3.0-1ubuntu2' g++='4:9.3.0-1ubuntu2' libc6-dev='2.31-0ubuntu9.9'
 RUN wget -O /usr/bin/opam https://github.com/ocaml/opam/releases/download/2.1.3/opam-2.1.3-i686-linux && chmod 755 /usr/bin/opam
 
 ENV OPAMROOT=/tmp
```
- The change for `apt` to `apt-get` is driven by the last comment in the `apt` manpage: https://manpages.debian.org/bullseye/apt/apt.8.en.html#SCRIPT_USAGE_AND_DIFFERENCES_FROM_OTHER_APT_TOOLS
- The versions of `gcc`, `g++` and `libc` (I had to change the package to `libc-dev` for `libc6-dev`) are chosen to match the build.robur.coop ones (https://builds.robur.coop/job/qubes-firewall/build/14878d91-62b2-4ad8-bde5-acb23f6c6575/f/system-packages). While on that I wonder if the selection of a specific version for `binutils` can also be useful (for `as`, `ld`, etc.).
- I split the line of default package vs version_specific pakages to help readability

So far I have the same sha256sum and future compilation could fails only when that specific versions are out of the ubuntu repositories so issues will be triggered :)

Does that sounds good to you @hannesm?

h01ger:
On Sun, Nov 13, 2022 at 03:56:35PM -0800, Hannes Mehnert wrote:
> As alternative we can push intermediate docker containers (somewhere?) and have them be used by the script, but then certainly two more issues are present: how to reproduce that container and who is in charge of updating it?

Debian developers do publish and update reproducible Debian images on docker.io
(which btw are also usable with podman).


-- 
cheers,
	Holger

 ⢀⣴⠾⠻⢶⣦⠀
 ⣾⠁⢠⠒⠀⣿⡁  holger@(debian|reproducible-builds|layer-acht).org
 ⢿⡄⠘⠷⠚⠋⠀  OpenPGP: B8BF54137B09D35CF026FE9D 091AB856069AAA1C
 ⠈⠳⣄

There are only two kinds of nazis: stupid ones and those without an excuse.
(Volker Strübing)

hannesm:
Thanks for your comment, @h01ger. Would you mind to point me to a "reproducible Debian image on docker.io"? Maybe we can base our builds on that?

Is there a story about installing system packages in a reproducible way? So, does such a Docker image contain a apt cache/database to install stuff from (since running `apt update` may result in installing some unknown version that may affect reproducibility)?

h01ger:
On Mon, Nov 14, 2022 at 02:55:32AM -0800, Hannes Mehnert wrote:
> Thanks for your comment, @h01ger. Would you mind to point me to a "reproducible Debian image on docker.io"? Maybe we can base our builds on that?
 
https://hub.docker.com/_/debian (worth reading in full) and
https://docker.debian.net/

> Is there a story about installing system packages in a reproducible way?

debrebuild from src:devscripts should be able to do that, 
though https://github.com/fepitre/debrebuild might be a better start.
(the latter - despite the name which is about to be changed - is not a 
fork but rather a rewrite in python.)


-- 
cheers,
	Holger

 ⢀⣴⠾⠻⢶⣦⠀
 ⣾⠁⢠⠒⠀⣿⡁  holger@(debian|reproducible-builds|layer-acht).org
 ⢿⡄⠘⠷⠚⠋⠀  OpenPGP: B8BF54137B09D35CF026FE9D 091AB856069AAA1C
 ⠈⠳⣄

When you’re used to privilege, equality feels like oppression.

hannesm:
Thanks again, @h01ger. I think there's some misunderstanding, so let me rephrase what I'm looking for:
1. we need a base image, thanks to reproducible Debian image on https://hub.docker.com/_/debian
2. we need some packages (such as gcc, libc-dev, ...) installed onto the base image and would like to ensure that the very same package is installed (since e.g. upgrading the C compiler will result in a different binary)
3. on top of that, we'll install opam and the qubes-mirage-firewall (we ensure ourselves by using a specific opam-repository commit to always get the same opam packages (tarballs))

To solve (2), I can see two ways: either (1) has a apt-cache and we do not need to run `apt update` or we provide a list with versions (and/or hashes) of debian packages to install. Do you know how other projects try to deal with (2), or have some ideas how to approach this?

hannesm:
Thanks for the additional hyperlinks @h01ger, especially the debian archive snapshots may be very suitable for our use.

h01ger:
On Mon, Nov 14, 2022 at 04:07:53AM -0800, Hannes Mehnert wrote:
> Thanks for the additional hyperlinks @h01ger, especially the debian archive snapshots may be very suitable for our use.

very happy to help and to see all this recent activity on qubes-mirage-firewall too! :thumbsup:


-- 
cheers,
	Holger

 ⢀⣴⠾⠻⢶⣦⠀
 ⣾⠁⢠⠒⠀⣿⡁  holger@(debian|reproducible-builds|layer-acht).org
 ⢿⡄⠘⠷⠚⠋⠀  OpenPGP: B8BF54137B09D35CF026FE9D 091AB856069AAA1C
 ⠈⠳⣄

The devel is in the details.

Szewcson:
Just an idea to consider. Maybe this will be worth of trying: https://www.chainguard.dev/unchained/introducing-wolfi-the-first-linux-un-distro

hannesm:
I tried the "let's use the system-packages from builds.robur.coop in the Dockerfile" with not much success, since builds.robur.coop uses an older ubuntu 20.04 and the packages are not available anymore.

I suggest we merge this PR as is, and then work on what @h01ger proposed:
- use a reproducible debian container
- use a package snapshot source from http://snapshot.notset.fr/archive/debian/ (or even better once snapshot.reproducible-builds.org is up that one)

That will as well require to use the same container for builds on builds.robur.coop (but that is not an issue, just a reminder). I'll open a separate issue for this.

PR #169, opened 2022-12-09
Improve workflow

inipfs:
- Print digest to summary dashboard
- Add workflow dispatch action

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,6 +6,7 @@ on:
   schedule:
     # Prime the caches every Monday
     - cron: 0 1 * * MON
+  workflow_dispatch:
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


hannesm:
Could you please elaborate a bit what this means "improve workflow"? What is the "summary dashboard"? I'm also not sure what "workflow dispatch action" is...

.github/workflows/main.yml
```diff
@@ -21,10 +22,14 @@ jobs:
...
-
-      - run: sh -exc 'if [ $(sha256sum dist/qubes-firewall.xen | cut -d " " -f 1) = $(grep "SHA2 last known" build-with-docker.sh | rev | cut -d ":" -f 1 | rev | cut -d "\"" -f 1 | tr -d " ") ]; then echo "SHA256 MATCHES"; else exit 42; fi'
+      - name: Build
+        run: ./build-with-docker.sh
 
+      - name: Verify
+        run: |
+          sh -exc 'if [ $(sha256sum dist/qubes-firewall.xen | cut -d " " -f 1) = $(grep "SHA2 last known" build-with-docker.sh | rev | cut -d ":" -f 1 | rev | cut -d "\"" -f 1 | tr -d " ") ]; then echo "SHA256 MATCHES"; else exit 42; fi'
+          echo '### sha256' >> $GITHUB_STEP_SUMMARY
+          sh -exc 'sha256sum dist/qubes-firewall.xen >> $GITHUB_STEP_SUMMARY'
```
hannesm, line 32: should this be first put into `$GITHUB_STEP_SUMMARY`? since the first command fails if the checksum does not match (and that should result in a CI failure). Should the sha256 computation be done only once and preserved in a variable, instead of computing it twice?

Issue #173, opened 2023-03-30
Qubes R4.1


bluesteal:
Does the following tutorial apply to Qubes R4.1?

https://roscidus.com/blog/blog/2016/01/01/a-unikernel-firewall-for-qubesos/#the-xen-virtual-network-layer

I don't see a "Proxy-VM" option in that version of Qubes.

Also, there is no option to set the kernel to 'mirage-firewall'

bluesteal:
ok it works now; the only thing needed is to get the qubes libraries installed (GUI, qrexec, qubesDB)

palainp:
Thanks for your questions. You'd follow the README instructions for newer Qubes. It should works for Qubes 4.1.

> ok it works now;

Does that statement is relevant in #172 and should close it?

> the only thing needed is to get the qubes libraries installed (GUI, qrexec, qubesDB)

Those are directly included in the unikernel during the compilation. You can check the fw logs in dom0, you should be able to see that qrexec v3 is used and Fw rules are updated (that would mean that qrexec and qubesDB are correctly "installed"). GUI was removed a couple of month ago.

bluesteal:
Thanks for your response. Can you link me for the README? and how can I check fw rules in dom0?

hannesm:
The README of this repository includes a section about how to deploy it: https://github.com/mirage/qubes-mirage-firewall#deploy

"How can I check fw rules in dom0" -- In dom0 you have the Qubes Manager that can display (and edit) firewall rules, see https://www.qubes-os.org/doc/firewall/ -- these rules are used by the qubes mirage firewall.

palainp:
Thanks @hannesm, and @bluesteal can check the vm logs as well reading the file `/var/log/xen/console/guest-mirage-fw.log` with `mirage-fw` the name of your mirage unikernel AppVM.
You should find some information like:
```
Solo5: Xen console: port 0x2, ring @0x00000000FEFFF000
            |      ___|
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
Solo5: Bindings version v0.7.3
Solo5: Memory map: 32 MB addressable:
Solo5:   reserved @ (0x0 - 0xfffff)
Solo5:       text @ (0x100000 - 0x319fff)
Solo5:     rodata @ (0x31a000 - 0x384fff)
Solo5:       data @ (0x385000 - 0x53ffff)
[...]

Issue #165, opened 2022-11-17
Reproducibilty considerations and tweaks


hannesm:
from #164, we suffer a bit that
- the ubuntu docker container is not reproducible per-se
- we use `apt update` leading to a random set of installed packages

To solve these issue, thanks to @h01ger we have some improvements (that should be done to (a) build-with-docker.sh and (b) builds.robur.coop):
- use a reproducible (debian) container from https://hub.docker.com/_/debian
- use a package snapshot (e.g. http://snapshot.notset.fr/archive/debian/20220523T033231Z/ instead of https://deb.debian.org/ for your apt sources.list) NB: snapshot.notset.fr will soon (tm) also be available as under snapshot.reproducible-builds.org but we are not there yet

hannesm:
We could as well use `orb` in the GitHub action runner and collect/upload the build information.

palainp:
Done with #175.

PR #175, opened 2023-04-18
update build script

palainp:
A bad hashsum was reported in the Qubes forum (https://forum.qubes-os.org/t/mirage-firewall-dispvm-template/17881/10). It seems that ubuntu has updated some packages a couple of weeks ago, I checked and found some difference in the output of github actions:
```
http://archive.ubuntu.com/ubuntu focal-updates/main amd64 libkrb5support0 amd64 1.17-6ubuntu4.2 [31.0 kB]
http://archive.ubuntu.com/ubuntu focal-updates/main amd64 libk5crypto3 amd64 1.17-6ubuntu4.2 [80.0 kB]
http://archive.ubuntu.com/ubuntu focal-updates/main amd64 libkrb5-3 amd64 1.17-6ubuntu4.2 [330 kB]
http://archive.ubuntu.com/ubuntu focal-updates/main amd64 libgssapi-krb5-2 amd64 1.17-6ubuntu4.2 [121 kB]
http://archive.ubuntu.com/ubuntu focal-updates/main amd64 libcurl3-gnutls amd64 7.68.0-1ubuntu2.16 [232 kB]
```
vs
```
http://archive.ubuntu.com/ubuntu focal-updates/main amd64 libkrb5support0 amd64 1.17-6ubuntu4.3 [31.1 kB]
http://archive.ubuntu.com/ubuntu focal-updates/main amd64 libk5crypto3 amd64 1.17-6ubuntu4.3 [80.2 kB]
http://archive.ubuntu.com/ubuntu focal-updates/main amd64 libkrb5-3 amd64 1.17-6ubuntu4.3 [330 kB]
http://archive.ubuntu.com/ubuntu focal-updates/main amd64 libgssapi-krb5-2 amd64 1.17-6ubuntu4.3 [121 kB]
http://archive.ubuntu.com/ubuntu focal-updates/main amd64 libcurl3-gnutls amd64 7.68.0-1ubuntu2.18 [233 kB]
```
This causes the github actions to fail due to bad resulting hashsum. Not sure why these updates have an impact on the resulting binary.

So I tried to update the build script in respect with #165 (and while at it, I also updated the opam version and the opam-repository commit), does that's what you mean @hannesm? I'm unsure if you have debian builds in ROBUR :(
@h01ger does the `echo` writing over `/etc/apt/source.list` sounds correct to you, or there is another/better way to do that?

As a final note, if this PR is merged, we probably also need to cut a release as now a salt script is comparing the hashsum of the release to the expected hashsum in `build_with_docker.sh` (thanks to @100111001: https://github.com/mirage/qubes-mirage-firewall/issues/173#issuecomment-1500458523)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,21 @@
 # Pin the base image to a specific hash for maximum reproducibility.
 # It will probably still work on newer images, though, unless an update
 # changes some compiler optimisations (unlikely).
-# ubuntu-20.04
-FROM ubuntu@sha256:b25ef49a40b7797937d0d23eca3b0a41701af6757afca23d504d50826f0b37ce
+# bookworm-slim
+FROM debian@sha256:07c6cb2ae86479dcc1942a89b0a1f4049b6e9415f7de327ff641aed58b8e3100
+# and set the package source to a specific release too
+RUN echo deb http://snapshot.notset.fr/archive/debian/20230418T024659Z bookworm main > /etc/apt/sources.list
 
-RUN apt update && apt install --no-install-recommends --no-install-suggests -y wget ca-certificates git patch unzip make gcc g++ libc-dev
-RUN wget -O /usr/bin/opam https://github.com/ocaml/opam/releases/download/2.1.3/opam-2.1.3-i686-linux && chmod 755 /usr/bin/opam
+RUN apt update && apt install --no-install-recommends --no-install-suggests -y wget ca-certificates git patch unzip bzip2 make gcc g++ libc-dev
+RUN wget -O /usr/bin/opam https://github.com/ocaml/opam/releases/download/2.1.4/opam-2.1.4-i686-linux && chmod 755 /usr/bin/opam
 
 ENV OPAMROOT=/tmp
 ENV OPAMCONFIRMLEVEL=unsafe-yes
 # Pin last known-good version for reproducible builds.
 # Remove this line (and the base image pin above) if you want to test with the
 # latest versions.
-RUN opam init --disable-sandboxing -a --bare https://github.com/ocaml/opam-repository.git#c9b2f766b7c7009be8cd68ac423d0d5b36044aca
-RUN opam switch create myswitch 4.14.0
+RUN opam init --disable-sandboxing -a --bare https://github.com/ocaml/opam-repository.git#28b35f67988702df5018fbf30d1c725734425670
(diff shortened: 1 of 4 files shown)


palainp:
Oh I think this is the wrong way to go (or at least not the correct package repository), the InRelease file has a `Valid-Until: Tue, 25 Apr 2023 02:11:25 UTC` entry I didn't noticed earlier, which makes the build fail (it's there to avoid obsolete packages, but since we're using the distribution to compile the unikernel, it should only matter if there's a security issue in the build suite: compiler, libraries, etc).

palainp:
The fix was an easy part :)

However, I now have a more general question about how we can use the reproducible Debian repositories, should we try to have an automatic default release process (e.g. 1/2 releases per year, just updating the hashes of the debian/opam-repository/head qubes-mirage-firewall repositories) + special releases in case of more urgent fixes (like last december)?

Dockerfile
```diff
@@ -1,19 +1,21 @@
...
-FROM ubuntu@sha256:b25ef49a40b7797937d0d23eca3b0a41701af6757afca23d504d50826f0b37ce
+# bookworm-slim
+FROM debian@sha256:07c6cb2ae86479dcc1942a89b0a1f4049b6e9415f7de327ff641aed58b8e3100
+# and set the package source to a specific release too
+RUN printf "deb [check-valid-until=no] http://snapshot.notset.fr/archive/debian/20230418T024659Z bookworm main" > /etc/apt/sources.list
 
-RUN apt update && apt install --no-install-recommends --no-install-suggests -y wget ca-certificates git patch unzip make gcc g++ libc-dev
-RUN wget -O /usr/bin/opam https://github.com/ocaml/opam/releases/download/2.1.3/opam-2.1.3-i686-linux && chmod 755 /usr/bin/opam
+RUN apt update && apt install --no-install-recommends --no-install-suggests -y wget ca-certificates git patch unzip bzip2 make gcc g++ libc-dev
+RUN wget -O /usr/bin/opam https://github.com/ocaml/opam/releases/download/2.1.4/opam-2.1.4-i686-linux && chmod 755 /usr/bin/opam
```
hannesm, line 10:
```suggestion
RUN wget -O /usr/bin/opam https://github.com/ocaml/opam/releases/download/2.1.5/opam-2.1.5-i686-linux && chmod 755 /usr/bin/opam
```

PR #180, opened 2023-08-23
Improved readme and saltstack script by adding more detailed installation instructions and removing hard coded templates

100111001:


palainp:
Thank you !

PR #181, opened 2023-09-10
Using too little RAM causes Mirage to stop working.

51lieal:
I have several VMs that use Mirage Firewall, and when it reaches up to three VMs, all VMs lose network connectivity. A Mirage restart is required. After increasing the RAM, I no longer experience any issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,8 @@ Run this command in dom0 to create a `mirage-firewall` VM using the `mirage-fire
 qvm-create \
   --property kernel=mirage-firewall \
   --property kernelopts='' \
-  --property memory=32 \
-  --property maxmem=32 \
+  --property memory=64 \
+  --property maxmem=64 \
   --property netvm=sys-net \
   --property provides_network=True \
   --property vcpus=1 \
--- a/SaltScriptToDownloadAndInstallMirageFirewallInQubes.sls
+++ b/SaltScriptToDownloadAndInstallMirageFirewallInQubes.sls
@@ -73,8 +73,8 @@ create-sys-mirage-fw:
       - kernel: mirage-firewall
       - kernelopts:
       - include-in-backups: False
-      - memory: 32
-      - maxmem: 32
+      - memory: 64
+      - maxmem: 64
       - netvm: sys-net
       - provides-network: True
       - vcpus: 1


palainp:
Thanks you for this PR. Does that memory issue appears with Qubes 4.2? I currently have no such issues with 4.1 even with 6+ appVM clients.

51lieal:
ah yes i was using qubes 4.2 sorry to mention that.
to give you better mapping this is my network route as example : 

sys-net > sys-mirage > mg-a (mirage) > sys-a
sys-net > sys-mirage > mg-b (mirage) > sys-b
sys-net > sys-mirage > mg-c (mirage) > sys-c

After 5 minutes of startup, every vm doesn't have connection, and i need to restart everything until the sys-net.

I have try without sys-mirage, and its still same, then i try to increase sys-mirage only to 64mb ram (mg-abc still 32mb) and everything is working. So, I think it's better to add more RAM to prevent unwanted issues in the future.

palainp:
Thanks for the explanation. As I'm unsure when I'll be able to update Qubes to 4.2 for testing further what's going on with memory, I think updating the installation instructions is the way to go.

Would you mind to also update the salt script with corresponding 64MB memory parameter?

51lieal:
Ah just by now, it only stand for 2h-3h until networking is lost. im now trying to increase mg-abc to 64mb ram, lets see how long it will work.

> Would you mind to also update the salt script with corresponding 64MB memory parameter?

I will send pr for that after this test, perhaps tomorrow to see if increasing mg-abc to 64mb will resolve the problem.

palainp:
Just out of curiosity, what is mg-[abc], another firewall VM or something else mirage-related (vpn router, dns-resolver, etc.)?

And you can update this PR with a new commit, you don't need to open a separate PR :)

51lieal:
Yeah, I can confirm that increasing the RAM on sys-mirage can resolve the problem. mg-abc is a Mirage Firewall; it's just the actual VM that is blocking websites. sys-mirage is meant for replacing sys-firewall without any rules.

hannesm:
Hmm, shouldn't we strive to figure out what is causing the memory usage, and fixing that? I think @xaki23 had quite some VMs connected to qmf, and only needs 42MB (is my memory correct)?

palainp:
Yes indeed, unfortunately at the moment I can't afford to update my day laptop to Qubes 4.2 because Qubes still in the RC state :(
But I'll update as soon as it will be stable enough and I'll investigate what's going on.

It seems that memory management has changed a bit in 4.2 (https://forum.qubes-os.org/t/memory-balloon-working-in-4-2/20380) and maybe what I've done in https://github.com/mirage/qubes-mirage-firewall/commit/b288481d2ffc7dc71f37db6aee515babf0dfa56e is not enough/the right way to do it. If Xen tries to get some memory back from Ocaml runtime bad things happen :(

Issue #184, opened 2023-11-07
How to get a reproducible build


ben-grande:
## Software version

I have built with docker on Debian using the docker repository
- Docker version 24.0.7, build afdd53b
- Debian 12
- 700e03de85c6914856e2e8c29fce267481f1e4e2 (HEAD commit)

## Logs

Build log:
> sha256sum mirage-firewall.tar.bz2 > mirage-firewall.sha256
SHA2 of build:   e5aeaee4f6b20b539565892da06361348a9bdca9b2fe3b29f5d2047e85d17ffc  ./dist/qubes-firewall.xen
SHA2 last known: 8ae5314edf5b863b788c4b873e27bc4b206a2ff7ef1051c4c62ae41584ed3e14
(hashes should match for released versions)

Manual check of build:
```sh
% sha256sum _build/mirage-firewall/vmlinuz
```
> e5aeaee4f6b20b539565892da06361348a9bdca9b2fe3b29f5d2047e85d17ffc  _build/mirage-firewall/vmlinuz

Manual check of release v0.8.5:
```sh
% sha256sum vmlinuz
```
> 8ae5314edf5b863b788c4b873e27bc4b206a2ff7ef1051c4c62ae41584ed3e14  vmlinuz

## Questions

I have read that docker should be run on Fedora, but I used the Docker repository on Debian, maybe it caused the difference?

What can I do to get a reproducible build?

palainp:
Hi @ben-grande, thanks for your detailed report.

The build host should not change the reproducibility result. However the Dockerfile assumes that http://snapshot.notset.fr/rchive/debian/ provides packets snapshots for debian (and is unfortunately offline so far, leading to `apt update` to use distributed debian packets). Therefore as the hashsum relies on the debian compilation toolchain stability and there was some update from time to time. I think this is the root cause for the hash difference. 

I tested and got the exact same hashsum both with fresh debian-12 and fedora-38 AppVMs. 

Now, to go further I changed Dockerfile snapshots from notset to debian, checked that every versions from debian packages are the same as in the last github action log, as well as every opam packages versions, and I still have the same hashsum as yours. I'll have to look deeper into packages versions and unfortunately I don't have anymore temporary compilation artifacts from a matching hash :(

palainp:
[...]

Issue #183, opened 2023-10-01
On fedora 38 template selinux preventing using docker with it root dir symlinked to dir in home folder


Szewcson:
When tried to follow build instruction of fedora38 based VM I encountered that error:
```
sudo ./build-with-docker.sh
Building Docker image with dependencies..
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/

Sending build context to Docker daemon  179.2kB
Step 1/13 : FROM debian@sha256:07c6cb2ae86479dcc1942a89b0a1f4049b6e9415f7de327ff641aed58b8e3100
 ---> cb5c02e18207
Step 2/13 : RUN printf "deb [check-valid-until=no] http://snapshot.notset.fr/archive/debian/20230418T024659Z bookworm main" > /etc/apt/sources.list
 ---> Running in 742b578c31d1
/bin/sh: error while loading shared libraries: /lib/x86_64-linux-gnu/libc.so.6: cannot apply additional memory protection after relocation: Permission denied
The command '/bin/sh -c printf "deb [check-valid-until=no] http://snapshot.notset.fr/archive/debian/20230418T024659Z bookworm main" > /etc/apt/sources.list' returned a non-zero code: 127

```

palainp:
Hi @Szewcson, thanks for the report. I'm surprised, to reproduce the issue I just installed a fresh fedora-38 template (just a couple of packages installed (patch, git, etc.)) and I don't suffer a similar issue:
```
$ sudo ./build-with-docker.sh 
Building Docker image with dependencies..
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/

Sending build context to Docker daemon  176.1kB
Step 1/13 : FROM debian@sha256:07c6cb2ae86479dcc1942a89b0a1f4049b6e9415f7de327ff641aed58b8e3100
docker.io/library/debian@sha256:07c6cb2ae86479dcc1942a89b0a1f4049b6e9415f7de327ff641aed58b8e3100: Pulling from library/debian
143db7d08421: Pull complete 
Digest: sha256:07c6cb2ae86479dcc1942a89b0a1f4049b6e9415f7de327ff641aed58b8e3100
[...]

PR #185, opened 2023-11-08
Fix docker build & update build scripts

palainp:
This PR wants to address #183 and #184:
- With Qubes 4.2 SELinux policies are enforced and now we need to run an additional command to be allowed to store docker images in the user homedir
- A recent update in opam-overlay changed the result hashsum, therefore I updated the build scripts (change from notset.fr to official debian repositories, update to mirage 4.4.0, ...)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,4 +26,7 @@ RUN opam exec -- opam install -y mirage opam-monorepo ocaml-solo5
 RUN mkdir /tmp/orb-build
 ADD config.ml /tmp/orb-build/config.ml
 WORKDIR /tmp/orb-build
-CMD opam exec -- sh -exc 'mirage configure -t xen --allocation-policy=best-fit && make depend && make tar'
+CMD opam exec -- sh -exc 'mirage configure -t xen --extra-repos=\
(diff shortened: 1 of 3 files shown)


hannesm:
thanks, looks good to me :)

palainp:
Thank you @hannesm!
I just try to pin the opam-overlay directory to fix the last reproducibility issue before merging.

hannesm:
FWIW, we can from a release link to the build-info from `https://builds.robur.coop/hash?sha256=<SHA256-of-the-unikernel-image>`. And hopefully in the future document how you can replay / rebuild it ;) (there's `orb rebuild` - but we do _not_ test this periodically -- it also lacks the information that you should run it in that specific docker container, and the apt/sources.list is not stored)

palainp:
So pinning opam-overlay to an earlier commit permits me to get the previous hash. I think the situation is better now.

palainp:
The last commit is to inform the user about the temporary disk usage with docker (thanks to @ben-grande for pointing out that in #186).

hannesm:
I do approve these changes. thanks a lot!

Issue #187, opened 2023-11-08
Improve build security


ben-grande:
## Problem

Currently, Mirage Firewall is aiming to reproducible builds. But can it also improve the build security?

I read the git blames but they didn't inform the implications of these lines, just why they were need, for reproducibility.

0. Can apt-transport-https be installed before setting the [debian snapshot](https://github.com/mirage/qubes-mirage-firewall/blob/700e03de85c6914856e2e8c29fce267481f1e4e2/Dockerfile#L7) so debian packages will be fetched over https for defense in depth against a failure of dpkg signature verification, which happens [from time to time](https://askubuntu.com/a/1114040). I am proposing an `apt upgrade` and `apt install -y --no-install-recommends apt-transport-tor` and change the repo definition of the snapshot to use `https`. Will this break reproducibility because this package will not be pinned to a specific debian snapshot?

1. [download of Opam without signature verification](https://github.com/mirage/qubes-mirage-firewall/blob/700e03de85c6914856e2e8c29fce267481f1e4e2/Dockerfile#L10), although the release packages have a companion `.sig` signature file. Would need to save the Opam release public key to this repo and place it in the docker container.

2. Why is [OPAMCONFIRMLEVEL=unsafe-yes](https://github.com/mirage/qubes-mirage-firewall/blob/700e03de85c6914856e2e8c29fce267481f1e4e2/Dockerfile#L13) require? What insecurities can it bring? I know this is Opam specific and not qubes-mirage-firewall, but I didn't understand the implications from reading the [manpage](https://opam.ocaml.org/doc/man/opam.html).

3. Why is [--disable-sandboxing](https://github.com/mirage/qubes-mirage-firewall/blob/700e03de85c6914856e2e8c29fce267481f1e4e2/Dockerfile#L17) needed? Is this to downgrade the opam version to be pinnned?

4. Is Opam fetching through https [0](https://github.com/mirage/qubes-mirage-firewall/blob/700e03de85c6914856e2e8c29fce267481f1e4e2/Dockerfile#L17) [1](https://github.com/mirage/qubes-mirage-firewall/blob/700e03de85c6914856e2e8c29fce267481f1e4e2/Dockerfile#L19) solely? This is also Opam specific, Opam does not do package signature verification? If it does not do, this is an upstream problem and serves only as an information for me as it can't be fixed here. I have not found information that Opam does software verifications besides the CA certificates.

## Summary

Analysis:

- 0: probably interferes with reproducible builds;
[...]

PR #190, opened 2023-12-26
check opam hashsum in Dockerfile

palainp:
This PR want to address some comments on #187 . To my understanding:
- 0: not solvable without a custom container that includes the `ca-certificates` package
- 1: is solved here
- 2, 3, and 4: answered in the issue comments.
/cc @ben-grande, @hannesm

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 # changes some compiler optimisations (unlikely).
 # bookworm-slim taken from https://hub.docker.com/_/debian/tags?page=1&name=bookworm-slim
 FROM debian@sha256:ea5ad531efe1ac11ff69395d032909baf423b8b88e9aade07e11b40b2e5a1338
-# install ca-certificates and remove default packages repository
+# install remove default packages repository
 RUN rm /etc/apt/sources.list.d/debian.sources
 # and set the package source to a specific release too
 # taken from https://snapshot.debian.org/archive/debian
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -13,6 +13,9 @@ RUN printf "deb [check-valid-until=no] http://snapshot.debian.org/archive/debian
 
 RUN apt update && apt install --no-install-recommends --no-install-suggests -y wget ca-certificates git patch unzip bzip2 make gcc g++ libc-dev
 RUN wget -O /usr/bin/opam https://github.com/ocaml/opam/releases/download/2.1.5/opam-2.1.5-i686-linux && chmod 755 /usr/bin/opam
+# taken from https://raw.githubusercontent.com/ocaml/opam/master/shell/install.sh
+RUN test `sha512sum /usr/bin/opam | cut -d' ' -f1` = \
+"38802b3079eeceb27aab3465bfd0f9f05a710dccf9487eb35fa2c02fbaf9a0659e1447aa19dd36df9cd01f760229de28c523c08c1c86a3aa3f5e25dbe7b551dd" || exit
```
kit-ty-kate, line 18: I’m not sure about this.

in bash `x || exit` always returns 0 as `exit` is `exit 0` by default, so this whole new checks does not do anything.

kit-ty-kate: Actually nevermind I should’ve read the manual page
> If n is omitted, the exit status is that of the last command executed.

Today I Learned. Sorry to have bothered you

palainp: Thanks for the double check!

Issue #132, opened 2021-02-24
Consider using podman instead docker


donob4n:
It seems that docker is getting abandoned in fedora in favor to podman so currently without passing some kernel args docker is even unusable in fedora-32.

I just tried installing it, 'sudo dnf install podman' and replacing 'docker' by 'podman' in 'build-with-docker.sh' and it worked!:

```
tar cjf mirage-firewall.tar.bz2 -C _build --mtime=./build-with-docker.sh mirage-firewall
SHA2 of build:   0991877acc49b1c55a57dfa50df157fab7cc6c2fce5727f605e64a0bc73d322e  qubes_firewall.xen
SHA2 last known: d68d2a8d2337b8c1a78995e1acbb4f72082076c73be45bf40dd6268c87b2353e

```
Notes:
- Run with sudo or it will fail due permissions errors
- Extend root partition or it ends due out of space.

palainp:
Thanks for your suggestion, and sorry for the delay. I so far have no issues with docker and f37. Using podman is currently not planed.

h01ger:
On Wed, Jul 05, 2023 at 10:11:04AM -0700, Pierre Alain wrote:
> Thanks for your suggestion, and sorry for the delay. I so far
> have no issues with docker and f37. Using podman is currently not planed.

sad face :(


-- 
cheers,
	Holger

 ⢀⣴⠾⠻⢶⣦⠀
 ⣾⠁⢠⠒⠀⣿⡁  holger@(debian|reproducible-builds|layer-acht).org
 ⢿⡄⠘⠷⠚⠋⠀  OpenPGP: B8BF54137B09D35CF026FE9D 091AB856069AAA1C
 ⠈⠳⣄

The law, in its majestic equality, forbids the rich as well as the poor to
sleep under bridges, to beg in the streets, and to steal bread. (Anatole France)

hannesm:
I'm sure, now with "debian 12" and the packages from notset.fr being in use, this can be achieved easily. So, a PR `./build-with-podman.sh` (optimally reproducing the same hash) and a CI job that executes it would be very welcome. It is just our development resources on qubes-mirage-firewall won't cover to use podman, but we're happy to review pull requests that contribute it (and to avoid bitrot, the CI job is pretty important).

h01ger:
On Thu, Jul 06, 2023 at 01:56:22AM -0700, Hannes Mehnert wrote:
[...]

PR #191, opened 2023-12-26
allow podman as building system

palainp:
This PR wants to go further with #132 by using the very same build script with docker and podman. We'll now have to workflows and the binary result should be the same.


palainp:
Big delay after the issue, @donob4n, @h01ger, I think this could solve #132.

PR #178, opened 2023-08-05
manage a dynamic uplink

palainp:
Dear developpers, this PR wants to solve #146 [1] and #156 (or at least helps a bit). To do this I had to merge several source files (uplink, client_net, router, firewall) into a single one.
I also had to add some mutable fields :( and haven't figured out how to do that another way so far.
With all theses changes around I'll feel way more confident if any Ocaml guru wants to look further into `dispatcher.ml`. There is certainly room for improvements here :)

[1]: I didn't managed to set up properly a OpenBSD netvm, but with this PR it correctly sends packets to the désigned gw IP.

PR #192, opened 2024-04-20
Update to Mirage 4.5.0

palainp:
Dear devs,
This PR wants to update to recent mirage API. In doing so it appears that `Memory.quick_stat()` reports wrong values (~~I haven't investigated why yet, and maybe it was present before, but now it goes lower than the limits in `memory_pressure.ml`~~ EDIT: There was a bad computation in `ocaml-solo5` again):
```
Memory.quick_stat() INF [memory_pressure] Writing meminfo: free 11MiB / 27MiB (39.99 %)
Memory.stat() INF [memory_pressure] Writing meminfo: free 19MiB / 27MiB (70.31 %)
```
~~With wrong values, the firewall starts to drop packets (`Memory_critical` being returned by `Memory_pressure.status()`). So I've replaced `Memory.quick_stat()` with `Memory.stat()`, but this is at the cost of going through the whole heap and `Memory_pressure.status` calls it for each packets. But I'd rather the firewall be slower than unusable :)~~

~~Going further will need to dive more into `ocaml-solo5/nolibc/malloc.i` as the bug seems to come from there but the code is really hard :'(~~
Best,

### DIFF
--- a/config.ml
+++ b/config.ml
@@ -1,3 +1,4 @@
+(* mirage >= 4.5.0 *)
 (* Copyright (C) 2017, Thomas Leonard <thomas.leonard@unikernel.com>
    See the README file for details. *)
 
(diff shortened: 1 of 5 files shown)


palainp:
With https://github.com/mirage/ocaml-solo5/pull/135 using `quick_stat()` is ok now so I removed the last modifications to have it back :)

palainp:
Will be included in #178 .

config.ml
```diff
@@ -1,37 +1,32 @@
+(* mirage >= 4.5.0 *)
```
hannesm, line 1: & < 5.0 (from what I'd expect) ;)

config.ml
```diff
@@ -1,37 +1,32 @@
...
       package "mirage-net-xen";
       package "ipaddr" ~min:"5.2.0";
       package "mirage-qubes" ~min:"0.9.1";
-      package ~min:"3.0.1" "mirage-nat";
+      package "mirage-nat" ~min:"3.0.1";
       package "mirage-logs";
       package "mirage-xen" ~min:"8.0.0";
-      package ~min:"6.4.0" "dns-client";
+      package "dns-client" ~min:"6.4.0";
       package "pf-qubes";
```
hannesm, line 30: are these argument reorderings needed?

memory_pressure.ml
```diff
@@ -26,6 +25,7 @@ let meminfo stats =
                   Cached: 0 kB\n\
                   SwapTotal: 0 kB\n\
                   SwapFree: 0 kB\n" (mem_total / 1024) (mem_free / 1024)
+*)
```
hannesm, line 28: let's remove unused code -- there's git for the archeologists

palainp:
Sorry I closed while you are reviewing it, I'll add your changes in #178.

hannesm:
@palainp thanks. no worries. :)

config.ml
palainp: I don't think it needed, this was just cosmetic to keep packages aligned.

config.ml
palainp: yes!

memory_pressure.ml
palainp: that's right :)

PR #193, opened 2024-04-24
drop astring dependency

hannesm:

PR #194, opened 2024-05-06
allow the firewall to use the router for dns requests (in rules)

palainp:
This PR fixes a DNS issue that was onboarded in the last release. The DNS client of the firewall was not correctly configured, leading qmf unable to resolve rules with domain names. (fix is confirmed by @neoniobium at https://forum.qubes-os.org/t/mirage-firewall-0-9-0-released/26030/11)
I didn't updated the hashsum in this PR as I think this PR and #193 can both be merged and then we can cut a release with the final hashsum :)


hannesm (approved):
looks good to me

hannesm:
I included your commit in #193 so we can have one run and get one checksum update :)

PR #197, opened 2024-05-22
Use Lwt.Syntax and avoid some >>= fun () patterns

dinosaure:
This little commit takes the advantage of `let*` to replace some parts of the code. In my opinion (feel free to take it in account), the code is more readable than before. About the `vifs` function, I take the advantage that `String.split_on_char` can not return an empty list according to the documentation: we can pattern-match with `[@warning "-8"]` or use `List.hd`.

### DIFF
--- a/dao.ml
+++ b/dao.ml
@@ -81,7 +81,7 @@ let vifs client domid =
           let get_client_ip () =
             let* str = Xen_os.Xs.read handle (Fmt.str "%s/%d/ip" path device_id) in
             let client_ip = List.hd (String.split_on_char ' ' str) in
-            (* XXX(dinosaure): it's safe to use [List.hd] here,
+            (* NOTE(dinosaure): it's safe to use [List.hd] here,
                [String.split_on_char] can not return an empty list. *)
             Lwt.return_some (vif, Ipaddr.V4.of_string_exn client_ip)
           in
--- a/dao.ml
+++ b/dao.ml
@@ -65,43 +65,35 @@ let read_rules rules client_ip =
                              number = 0;})]
 
 let vifs client domid =
+  let open Lwt.Syntax in
   match int_of_string_opt domid with
   | None -> Log.err (fun f -> f "Invalid domid %S" domid); Lwt.return []
   | Some domid ->
-    let path = Printf.sprintf "backend/vif/%d" domid in
-    Xen_os.Xs.immediate client (fun handle ->
-        directory ~handle path >>=
-        Lwt_list.filter_map_p (fun device_id ->
-            match int_of_string_opt device_id with
-            | None -> Log.err (fun f -> f "Invalid device ID %S for domid %d" device_id domid); Lwt.return_none
-            | Some device_id ->
-              let vif = { ClientVif.domid; device_id } in
-              Lwt.try_bind
-                (fun () -> Xen_os.Xs.read handle (Printf.sprintf "%s/%d/ip" path device_id))
-                (fun client_ip ->
-                   let client_ip' = match String.split_on_char ' ' client_ip with
-                     | [] -> Log.err (fun m -> m "unexpected empty list"); ""
-                     | [ ip ] -> ip
-                     | ip::rest ->
-                       Log.warn (fun m -> m "ignoring IPs %s from %a, we support one IP per client"
-                                    (String.concat " " rest) ClientVif.pp vif);
-                       ip
-                   in
-                   match Ipaddr.V4.of_string client_ip' with
-                   | Ok ip -> Lwt.return (Some (vif, ip))
-                   | Error `Msg msg  ->
-                     Log.err (fun f -> f "Error parsing IP address of %a from %s: %s"
-                                 ClientVif.pp vif client_ip msg);
-                     Lwt.return None
-                )
-                (function
-                  | Xs_protocol.Enoent _ -> Lwt.return None
-                  | ex ->
-                    Log.err (fun f -> f "Error getting IP address of %a: %s"
-                                ClientVif.pp vif (Printexc.to_string ex));
-                    Lwt.return None
-                )
-          ))
+    let path = Fmt.str "backend/vif/%d" domid in
+    let fn handle =
+      let* entries = directory ~handle path in
+      let fn device_id = match int_of_string_opt device_id with
+        | None ->
+          Log.err (fun f -> f "Invalid device ID %S for domid %d" device_id domid);
+          Lwt.return_none
+        | Some device_id ->
+          let vif = { ClientVif.domid; device_id } in
+          let fn () =
+            let* str = Xen_os.Xs.read handle (Fmt.str "%s/%d/ip" path device_id) in
+            let[@warning "-8"] client_ip :: _ = String.split_on_char ' ' str in
+            Lwt.return_some (vif, Ipaddr.V4.of_string_exn client_ip) in
+          Lwt.catch fn @@ function
+          | Xs_protocol.Enoent _ -> Lwt.return_none
+          | Ipaddr.Parse_error (msg, client_ip) ->
+            Log.err (fun f -> f "Error parsing IP address of %a from %s: %s"
+              ClientVif.pp vif client_ip msg);
+            Lwt.return_none
+          | exn ->
+            Log.err (fun f -> f "Error getting IP address of %a: %s"
+              ClientVif.pp vif (Printexc.to_string exn));
+            Lwt.return_none in
+      Lwt_list.filter_map_p fn entries in
+    Xen_os.Xs.immediate client fn
 
 let watch_clients fn =
   Xen_os.Xs.make () >>= fun xs ->
--- a/dao.ml
+++ b/dao.ml
@@ -81,7 +81,8 @@ let vifs client domid =
           let get_client_ip () =
             let* str = Xen_os.Xs.read handle (Fmt.str "%s/%d/ip" path device_id) in
             let client_ip = List.hd (String.split_on_char ' ' str) in
-            Lwt.return_some (vif, Ipaddr.V4.of_string_exn client_ip) in
+            Lwt.return_some (vif, Ipaddr.V4.of_string_exn client_ip)
+          in
           Lwt.catch get_client_ip @@ function
           | Xs_protocol.Enoent _ -> Lwt.return_none
           | Ipaddr.Parse_error (msg, client_ip) ->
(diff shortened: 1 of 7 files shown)


dao.ml
```diff
@@ -65,43 +65,35 @@ let read_rules rules client_ip =
...
-                  | ex ->
-                    Log.err (fun f -> f "Error getting IP address of %a: %s"
-                                ClientVif.pp vif (Printexc.to_string ex));
-                    Lwt.return None
-                )
-          ))
+    let path = Fmt.str "backend/vif/%d" domid in
+    let fn handle =
+      let* entries = directory ~handle path in
+      let fn device_id = match int_of_string_opt device_id with
```
hannesm, line 75: to avoid confusion, could we name the inner `fn` to something else?

dao.ml
```diff
@@ -65,43 +65,35 @@ let read_rules rules client_ip =
...
+        | None ->
+          Log.err (fun f -> f "Invalid device ID %S for domid %d" device_id domid);
+          Lwt.return_none
+        | Some device_id ->
+          let vif = { ClientVif.domid; device_id } in
+          let fn () =
```
hannesm, line 81: could we rename this `fn` as well to something else, maybe even what it does?

dao.ml
```diff
@@ -65,43 +65,35 @@ let read_rules rules client_ip =
...
+            let* str = Xen_os.Xs.read handle (Fmt.str "%s/%d/ip" path device_id) in
+            let[@warning "-8"] client_ip :: _ = String.split_on_char ' ' str in
```
hannesm, line 83: this removed the warning "ignoring IPs %s from %a, we support one IP per client", I'm not sure whether this warning has been observed in the wild...

hannesm:
looks pretty good, some minor remarks inline

dao.ml
palainp: I just grep'd my logs and there is no occurences :)
I'm only worried about adding of `[@warning "-8"]` (due to my Ocaml ignorance) and I would prefer a pattern matching structure. But that's just a personal view and I'll be ok with that :)

dao.ml
```diff
@@ -65,43 +65,35 @@ let read_rules rules client_ip =
...
+          | Xs_protocol.Enoent _ -> Lwt.return_none
+          | Ipaddr.Parse_error (msg, client_ip) ->
+            Log.err (fun f -> f "Error parsing IP address of %a from %s: %s"
+              ClientVif.pp vif client_ip msg);
+            Lwt.return_none
+          | exn ->
+            Log.err (fun f -> f "Error getting IP address of %a: %s"
+              ClientVif.pp vif (Printexc.to_string exn));
+            Lwt.return_none in
+      Lwt_list.filter_map_p fn entries in
```
palainp, line 95: To me, adding the `in` on the next line, correctly indented, helps the code reading :)

dinosaure:
Just to inform you, I continue to look deeper into the code and see many places where we can improve/upgrade the code :+1:. I will continue to propose some PRs about that: don't think that this is the only and last PR :+1:.

dao.ml
```diff
@@ -80,7 +80,7 @@ let vifs client domid =
           let vif = { ClientVif.domid; device_id } in
           let get_client_ip () =
             let* str = Xen_os.Xs.read handle (Fmt.str "%s/%d/ip" path device_id) in
-            let[@warning "-8"] client_ip :: _ = String.split_on_char ' ' str in
+            let client_ip = List.hd (String.split_on_char ' ' str) in
```
palainp, line 83: And maybe now, it's worth adding a comment explaining that we keep only the first address?

palainp:
> Just to inform you, I continue to look deeper into the code and see many places where we can improve/upgrade the code 👍. I will continue to propose some PRs about that: don't think that this is the only and last PR 👍.

Or continue into this one? :D

dinosaure:
> Or continue into this one? :D

As you want if it's clearer for you.

hannesm:
I think due to the checksum / release stuff, there are two or three options (as I don't quite know about the salt stuff):
- continue refactorings in this PR (and then maybe squash the existing commits into a single one)
- merge to a `dev` branch, which at the end we'll merge into the `main` branch and cut a release
- fix the salt script to use the latest release instead of the main branch (if this is what is happening)

It would be great to keep the checksum as produced by CI and the one in the build-with file in sync... so I guess the first option is a good one :)

hannesm:
Of course, it is as well fine to accumulate commits here and then do a squash-merge :D

palainp:
> I think due to the checksum / release stuff, there are two or three options (as I don't quite know about the salt stuff):
>     * continue refactorings in this PR (and then maybe squash the existing commits into a single one)
>     * merge to a `dev` branch, which at the end we'll merge into the `main` branch and cut a release
>     * fix the salt script to use the latest release instead of the main branch (if this is what is happening)
> 

I think solution (a) is the easiest so far, and (c) should be done at some point, and maybe it's easier to include the vmlinuz hashsum in the `.sha256` file in the release (but TBH I'm unsure checking the vmlinuz is mandatory if the `.bz2` hash already match?, and https://github.com/unman/shaker/tree/main/mirage only check the `.bz2` sum).

> It would be great to keep the checksum as produced by CI and the one in the build-with file in sync... so I guess the first option is a good one :)

I also agree that having the correct hashsum in `build-with.sh` is important to avoid weekly CI failures :)

dinosaure:
My last patch can be a bit controversial and we should stop at this point to see if the unikernel works again. I explained in the commit why I did this change. Let me know if it's ok or not.

palainp:
> My last patch can be a bit controversial and we should stop at this point to see if the unikernel works again. I explained in the commit why I did this change. Let me know if it's ok or not.

Good news, at this point, qmf is still working (not a deep test: 3 clients in and out, bandwidth speed is the same as before). I'm probably ok with that change (at least I still understand the code), but I'm interested with others' thoughts :)

dispatcher.ml
```diff
@@ -404,46 +403,54 @@ struct
         (function Lwt.Canceled -> Lwt.return_unit | e -> Lwt.fail e)
     in
     Cleanup.on_cleanup cleanup_tasks (fun () -> Lwt.cancel listener);
-    Lwt.pick [ qubesdb_updater; listener ]
+    (* XXX(dinosaure): [qubes_updater] and [listener] can be forgotten, our [cleanup_task]
```
hannesm, line 406: I'm not sure I understand the comment, usually when I read `XXX` there's something left to do? To me, it sounds more like a description of the (intricate) task semantics. So, maybe remove the `XXX`?

dispatcher.ml
```diff
@@ -404,46 +403,54 @@ struct
...
-          (fun ex ->
-            Log.warn (fun f ->
-                f "Error with client %a: %s" Dao.ClientVif.pp vif
-                  (Printexc.to_string ex));
-            Lwt.return_unit));
-    cleanup_tasks
+    let* () =
+      Lwt.catch (add_vif get_ts vif dns_client dns_servers ~client_ip ~router
+          ~cleanup_tasks qubesDB)
+      @@ fun exn ->
```
hannesm, line 421: I personally struggle with `@@` and find parenthesis much easier to read (since I'm always unsure how strong the `@@` binding strength is.

dispatcher.ml
```diff
@@ -404,46 +403,54 @@ struct
...
-               if not (Dao.VifMap.mem key !clients) then (
-                 let cleanup =
-                   add_client get_ts dns_client dns_servers ~router key ip_addr
-                     qubesDB
-                 in
-                 Log.debug (fun f -> f "client %a arrived" Dao.ClientVif.pp key);
-                 clients := !clients |> Dao.VifMap.add key cleanup)))
+    let open Lwt.Syntax in
+    let clients : Cleanup.t Dao.VifMap.t ref = ref Dao.VifMap.empty in
+    Dao.watch_clients @@ fun new_set ->
```
hannesm, line 433: see above about `@@`.

hannesm:
I like the last commit, thanks @dinosaure.

It would make reviewing easier if you could separate syntactic changes (using Lwt.Syntax, using `@@`) and semantic changes.

I added some comments, and thanks to @palainp for testing this. It is intricate code where we should ensure to test extensively before merging (with adding and removing clients). Maybe I'm also afraid since I spent quite some time on that code, and was sometimes in stuck states... But as said, I reviewed the code and it looks good.

Should we adapt ocamlformat to this repository, so we don't have to argue about indentation, where to put parenthesis and where to put the `in`?

palainp:
Unfortunately I don't have yet a deep test protocol, all is done manually, and I may miss something, but I'll try to write a testbed script soon :)

> Should we adapt ocamlformat to this repository, so we don't have to argue about indentation, where to put parenthesis and where to put the in?

Yes I think so :)

palainp:
I wrote a simple naive script (using 10 AppVM clients with 2G memory) that returned me some issue (certainly present before this PR). I got when changing the netvm property (maybe a packet is still pending in the ring when the unikernel tries to connect to a new netvm?):
```
[2024-05-23 10:31:14] 2024-05-23T08:31:14-00:00: [ERROR] [dispatcher] uncaught exception trying to send to uplink: 
[2024-05-23 10:31:14]                                                 Failure("Shared_page_pool.use after shutdown")
```
I also got in dom0 console:
```
2024-05-23 10:27:08.738 qrexec-client[58505]: process_io.c:188:process_io: vchan connection closed early (fds: 1 -1 -1, status: -1 -1)
```
but that's more related to my script to be to eager when connecting/disconnecting clients AppVMs. I'll have to rewrite that before sharing :)

dispatcher.ml
dinosaure: I can change to `NOTE`maybe? My rules is:
- `XXX` for a informal comment
- `NOTE` for a formal comment
- `TODO` for a todo

hannesm:
I'd like to ask what to do here. I'm happy to rebase this PR and if @palainp tests we merge it!? It'd esp. be wonderful to have it included before the next release, so we get something out of the nice cleanups here :) Unfortunately it conflicts now that #201 was merged.

hannesm:
I rebased on top of main, cleanup up a whitespace and revised XXX to NOTE.

The leftover comment is about "@@", but a `git grep "@@"` results in several hits, so let's leave that coding style for a separate PR.

palainp:
Thank you! I need to test it, and think this PR will be good to merge!

palainp (approved):
LGTM !

PR #199, opened 2024-08-09
update opam repository in Dockerfile

hannesm:
the reason behind this is that in the earlier commit, some urls point to unavailable urls.


palainp:
Thank you @hannesm , it works as expected and this PR LGTM. I plan to cut a release soon for mirage-qubes, and update the unikernel against that new release before cutting a new qubes-mirage-firewall release. In the meantime, I've opened #200.

PR #201, opened 2024-10-14
update to mirage 4.8

hannesm:
as advertised. ~~will need to update the checksum~~


palainp:
Thank you @hannesm for that update (I also reproduce the shasum locally)! I only have a minor remark.

palainp (approved):

palainp:
I'd like to merge it like this, thank you @hannesm, and wait a bit for a release. When I'll have more time I'll work on the salt script and cut a release :)

PR #203, opened 2024-10-15
update opam repository commit to use solo5 0.9 and mirage 4.8.1

hannesm:


palainp:
Thank you! While updating things, do you think it's worth updating the debian docker image, debian snapshots, and opam-overlays as well?

hannesm:
> Thank you! While updating things, do you think it's worth updating the 
> debian docker image

I don't know

> debian snapshots,

I don't know

> and opam-overlays as well?

there's been no change as far as I can tell.

palainp:
I don't know neither about the debian ones and I haven't heard anything critical about it, so I'm keen to keep my caches without having to download too many things :)

For opam-overlays, there were lwt_ppx and something about colombe since the current commit hash, but I'm unsure if we need the updates or not.

hannesm:
we don't ;)

palainp:
Sorry for the force push, I had to remove the last commit to be able to merge, and I'll reintroduce the updated hashsum value on main.

Issue #200, opened 2024-08-10
Salt script broken


palainp:
As `initramfs` is no longer needed (see #70) we will be able to release only the unikernel file. With that in mind, the salt script can be simplified (no more tar involved) and, in the process, we can try to check the hashsum against something other than the one in the build script (maybe the best option would be to ship a hashsum file along the unikernel file?).

This will allow PRs to be merged (like #199) without needing for an immediate release matching the hashsum (and that explains this issue's title :) ).

palainp:
While preparing the next release, I wonder if a possible solution would be to have:
- a `qubes-firewall.sha256` file containing the hashsum of the current head (the one used in GH actions)
- a `qubes-firewall-release.sha256` file containing the hashsum of the latest release (the one used in the salt script)
- releases files will contain `vmlinuz` and `qubes-firewall-release.sha256`

Does anyone have a idea about that?

palainp:
Done (or will be done) in #205

PR #205, opened 2024-10-17
Update the salt script + releases files

palainp:
This PR wants to solve #200 and emphasises the separation of sha256 for released versions and for the current version of head. This way we can push commits on top of head, continue to ensure reproducible builds, and also keep something to compare the shasum of the released version. The PR process will need updating `qubes-firewall.sha256`, and the release process need to copy this file into `qubes-firewall-release.sha256`. WDYT?

I'm unsure whether the salt script has been updated correctly and I'll have to wait for the next version to fix any issues.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -23,7 +23,7 @@ jobs:
 
       - run: ./build-with.sh docker
 
-      - run: sh -exc 'if [ $(sha256sum dist/qubes-firewall.xen) = $(cat qubes-firewall.sha256) ]; then echo "SHA256 MATCHES"; else exit 42; fi'
+      - run: sh -exc 'if [ "$(sha256sum dist/qubes-firewall.xen)" = "$(cat qubes-firewall.sha256)" ]; then echo "SHA256 MATCHES"; else exit 42; fi'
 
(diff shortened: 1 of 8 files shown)


hannesm:
I think the approach is fine.

palainp:
I still have an issue the first time I run the script, because the directory for the kernel is not yet created. So far I'm unsure if I have a bad usage of salt or if just need to create the directory if `test -e` fails on the release version check:
```bash
$ sudo qubesctl --show-output state.apply SaltScript...
[ERROR   ] Command 'test' failed with return code: 1
[ERROR   ] stdout: touch: cannot touch '/var/lib/qubes/vm-kernels/mirage-firewall/version.txt': No such file or directory
cat: /var/lib/qubes/vm-kernels/mirage-firewall/version.txt: No such file or directory
[ERROR   ] retcode: 1
[ERROR   ] Command 'test' failed with return code: 1
[ERROR   ] output: touch: cannot touch '/var/lib/qubes/vm-kernels/mirage-firewall/version.txt': No such file or directory
cat: /var/lib/qubes/vm-kernels/mirage-firewall/version.txt: No such file or directory
```

Issue #208, opened 2024-12-19
Race condition when a client disconnect?


palainp:
Dear developpers,

On the Qubes forum, a user has a special setting, qubes-mirage-firewall uses OpenBSD (the unikernel is configured manually with no netvm, as described in README.md), and the latest release prevent that setup to run correctly.

To my understanding, the netvm client connect twice to the unikernel (not sure if that is mandatory, but it worked earlier), and in the logs, the logs lines are (this is the extract that seems relevant, I anonymized manually, XX=YY+1):
```
...
[INFO] [dispatcher] add client vif {domid=XX;device_id=0} with IP A.B.C.D
[INFO] [dispatcher] Client XX (IP: A.B.C.D) ready
[INFO] [ethernet] Connected Ethernet interface fe:ff:ff:ff:ff:ff
[INFO] [dispatcher] New firewall rules for A.B.C.D
0   any   accept
[INFO] [dispatcher] add client vif {domid=YY;device_id=0} with IP A.B.C.D
[INFO] [qubes.db] got rm "/qubes-firewall/A.B.C.D/"
[INFO] [qubes.db] got update: "/qubes-firewall/A.B.C.D/policy" = "drop"
[INFO] [qubes.db] got update: "/qubes-firewall/A.B.C.D/0000" = "action=accept"
[INFO] [qubes.db] got update: "/qubes-firewall/A.B.C.D" = ""
[INFO] [dispatcher] Rules did not change for A.B.C.D
[INFO] [net-xen backend] Frontend asked to close network device dom:XX/vif:0
[INFO] [dispatcher] Client YY (IP: A.B.C.D) ready
[INFO] [ethernet] Connected Ethernet interface fe:ff:ff:ff:ff:ff
[INFO] [dispatcher] New firewall rules for A.B.C.D
0   any   accept
[domYY:A.B.C.D] [client_eth] Waiting for old client domXX:A.B.C.D to go away before accepting new on
...
```
So client XX connects and shutdown (again, maybe an artifact due to the way OpenBSD connects as netvm), and client YY cannot connect after client XX has shut down.

The behaviour with previous release is:
```
...
[INFO] [dispatcher] add client vif {domid=XX;device_id=0} with IP A.B.C.D
[INFO] [dispatcher] Client XX (IP: A.B.C.D) ready
[INFO] [ethernet] Connected Ethernet interface fe:ff:ff:ff:ff:ff
[...]

PR #209, opened 2024-12-31
Fix openbsd as client

palainp:
As reported in #208 there is an issue with OpenBSD as sys-net. This PR fixes the issue with a minimal change to HEAD. I can squash the commits together (e.g. 1 for the modifications and 1 for the hashsum) to avoid multiple WIP commits, but as I referenced some in the issue, I keeped them so far.

### DIFF
--- a/dispatcher.ml
+++ b/dispatcher.ml
@@ -342,7 +342,7 @@ struct
 
   (** Connect to a new client's interface and listen for incoming frames and firewall rule changes. *)
   let conf_vif get_ts vif backend client_eth dns_client dns_servers
-      ~client_ip ~iface ~router ~cleanup_tasks qubesDB =
+      ~client_ip ~iface ~router ~cleanup_tasks qubesDB () =
     let { Dao.ClientVif.domid; device_id } = vif in
     Log.info (fun f ->
         f "Client %d:%d (IP: %s) ready" domid device_id (Ipaddr.V4.to_string client_ip));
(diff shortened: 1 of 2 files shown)


dispatcher.ml
```diff
@@ -427,6 +426,7 @@ struct
                 f "Error with client %a: %s" Dao.ClientVif.pp vif
                   (Printexc.to_string ex));
             Lwt.return_unit)) ;
+    Cleanup.on_cleanup cleanup_tasks (fun () -> remove_client router iface);
```
dinosaure, line 429: Actually, it will be preferable to set the clean-up **before** we launch the task to be sure that in any case, a finaliser exists to clean-up the added client.

palainp: I wasn't sure, and I moved around a lot on that, but ok to remove the last commit :)

palainp: Thank you !

dinosaure (approved):
For me it's ok if it's fix the initial issue reported. As far as a finalizer exists near the `Lwt.async`, I'm ok with that. The diffi is a bit noisy but it's ok.

If we look into details, we probably just need `Lwt.catch`. The initial issue is probably related to the fact that `add_client` raises an exception. Due to the fact that we did not catch it before this PR, the exception is propagated on top and fails our unikernel.

But let's fix it and release it as soon as we can :+1:

palainp:
Thank you !
I'll release in the current state, and we can update the code later :)

Issue #212, opened 2025-02-05
Netvm mac address when netvm doesn't reply


palainp:
Dear all,
A user on the Qubes forum tried to connect qubes-mirage-fw to a mullvad AppVM (mirage netvm is mullvad). There is an issue because qubes-mirage-fw cannot get the mac address of the VPN AppVM.
Interestingly, everything works fine with WireGuard, and as well mullvad AppVM works fine with the default linux sys-fw.
After some investigations it appears that the linux fw has associated a mac address for mullvad `fe:ff:ff:ff:ff:ff`.
I wonder if there isn't a potential problem in falling back on a mac broadcast address in case the netvm doesn't reply?

palainp:
It may be possible to add that "default mac" around https://github.com/mirage/qubes-mirage-firewall/blob/d8871f68c006ea7762323ededeec80fadc3ee178/dispatcher.ml#L574

palainp:
Solved by #213

PR #213, opened 2025-02-06
Fallback netvm mac

palainp:
Dear developpers,
This PR is what I had in mind to solve #212 , the fix is confirmed by the user that reported the issue with mullvad.
When searching in the QubesOS repository I got several occurrences of that precise mac address (see https://github.com/search?q=org%3AQubesOS%20fe%3Aff%3Aff%3Aff%3Aff%3Aff&type=code).

Just in case /cc @marmarek or @DemiMarie to confirm there is no security issue to fall back to that mac address in case our netvm does not reply to arp request ?

### DIFF
--- a/dispatcher.ml
+++ b/dispatcher.ml
@@ -571,7 +571,14 @@ struct
     I.connect ~cidr ~gateway eth arp >>= fun ip ->
     U.connect ip >>= fun udp ->
     let netvm_mac =
-      Arp.query arp gateway >|= or_raise "Getting MAC of our NetVM" Arp.pp_error
+      Arp.query arp gateway >>= function
+        | Error e ->
+          Log.err(fun f -> f "Getting MAC of our NetVM: %a" Arp.pp_error e);
+          (* This mac address is a special address used by Qubes when the device
+             is not managed by Qubes itself. This can occurs inside a service
+             AppVM (e.g. VPN) when the service creates a new interface. *)
+          Lwt.return (Macaddr.of_string_exn "fe:ff:ff:ff:ff:ff")
+        | Ok mac -> Lwt.return mac
     in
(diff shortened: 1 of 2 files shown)


palainp:
Sorry for the lack of context, a user experienced issues running a VPN vm and using qubes-mirage-firewall to filter traffic before the flow enters the VPN. The situation is `sys-net <- sys-fw <- sys-vpn <- qmf <- appvm`. If the user replaces qmf with a linux fw, all is fine. It looks like the vpn has a mac address `fe:ff:ff:ff:ff:ff` in the arp table of the linux fw.

Looking at the commit message from [network/setup-ip](https://github.com/QubesOS/qubes-core-agent-linux/commit/41fff23c6dae16e6eb3f686a79297f1fc5cd8ca7) this address is added permanently to avoid arp failures when the netvm doesn't respond to arp requests.

So I think we can either add it permanently or use this address as a fallback address, does anyone have a preference?

hannesm:
I do like the code as you suggest -- i.e. if the netvm replies to arp, use that mac address -- and otherwise use the hardcoded one provided here. thanks!

palainp:
Thank you ! So I merge and I'll cut a release soon as it may unblock other users too.

palainp:
Thinking twice, and after a bit of digging, all our client/netvm interfaces mac are `fe:ff:ff:ff:ff:ff:ff`, the only one different is ours :).
So I'm wondering if a better solution wouldn't be to always use the Qubes proxy method and always treat the remote mac as `fe:ff:ff:ff:ff:ff:ff:ff` without ARP request at all. Basically, we never have to worry about L2 packets and the NAT table will always tell us where to route the packets.
What do you think?

PR #215, opened 2025-03-10
update to mirage 4.9.0

hannesm:
this is mainly for discussion. There's no code reformatting in this PR -- I suggest that we start to use ocamlformat in this project (maybe similar to how https://github.com/robur-coop/mollymawk does it, with a GitHub action that pushes formatting changes).

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,45 @@
+name: ocamlformat
+
+on: [pull_request]
+
+jobs:
+  format:
+    name: ocamlformat
+
+    strategy:
+      fail-fast: false
+      matrix:
+        ocaml-version: ["4.14.2"]
+        operating-system: [ubuntu-latest]
+
+    runs-on: ${{ matrix.operating-system }}
+
+    steps:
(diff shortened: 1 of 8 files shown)


palainp:
Thank you! I think autoformat the code on push will be good, do you want to add it here (and make this PR an ecosystem update) or in a separate PR?

In the near future, I also want to update with a branch that onboard Ocaml5.3 (once new ocaml-solo5 is released) and will ask the Qubes forum to try out both versions (this PR Ocaml 4.14.2, and with 5.3) before milestone the 5+ releases. Then new features can be explored and added :) What do you think?

palainp:
Hmm, my ocamlformat addition is not quite right :( I'll dig into that deeper later.

hannesm:
@palainp exactly. for any PR where ocamlformat has a different formatting, the github action will push a commit. IMHO that's convenient (not nagging the author to "please run dune build @fmt".

palainp:
Thank you for your reply! So I can merge that PR since the two others checks are green, and admire the new commit once merged? Nice!

PR #217, opened 2025-03-17
fix formatting action

hannesm:


palainp:
\o/ now the format commit is put directly on top of the remote fork (for each push action), so I think this is a good way to add the ocamlformat action. WDYT?

PR #218, opened 2025-03-17
minor change

hannesm:
This is a test for the formatting CI action, where the base repository is the same (mirage/qubes-mirage-firewall), not a fork thereof.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ It's OK to install the Docker or Podman package in a template VM if you want it 
 after a reboot, but the build of the firewall itself should be done in a regular AppVM.
 
 You can also build without that script, as for any normal Mirage unikernel;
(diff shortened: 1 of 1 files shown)


hannesm:
ok, so the CI action works fine after this PR if we do not use a clone, but a branch in this repository directly.

palainp (approved):
Thank you @hannesm , I think we can use this updated CI, and updating with branches created here and not in the clone will be necessary from time to time until there is another CI update that allows adding the commit to external branches?

palainp:
I think that PR only changes README.md now ? Is that ok if I remove the other commits ?

hannesm:
> I think that PR only changes README.md now ? Is that ok if I remove the other commits ?

sure, this was only a test. we can as well close it.

palainp:
Thank you!

PR #219, opened 2025-07-15
Fix HVM clients AppVMs

palainp:
This PR wants to handle the case of HVM AppVMs.

HVM clients have two clients interfaces (appvm and appvm-dm), and therefore the unikernel see two vif interfaces. But only one should be active at the same time (one in state 4 "initialized" and the other in state 2 "initializing") and which one depends on the drivers of the system inside the AppVM. So to handle all cases we perform an initialization for both, but the client driver should theoritically only complete the initialization process for one of them (and here we catch the exception for the other).

### DIFF
--- a/dispatcher.ml
+++ b/dispatcher.ml
@@ -402,13 +402,14 @@ let conf_vif get_ts vif backend client_eth dns_client dns_servers ~client_ip
 
 (** A new client VM has been found in XenStore. Find its interface and connect
     to it. *)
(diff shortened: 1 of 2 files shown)


palainp:
The second series of commit are still needed to permit new clients even when one has not completed the init of its vif... Maybe that async can be racy bacause of line (`clients := Dao.VifMap.add key cleanup_tasks !clients;`)?

palainp:
This PR seems to run fine, at least "on my laptop" (tm). So merging and wait for breaking issues :)

